PR #1, opened 2020-08-23
Setup init workspace

cristian-aldea:
**NOTES ON REVIEWS: The PR does various things, so have a look at the commit messages for details. I would recommend going through the changes commit by commit and ignoring the commits "Add nx workspace", "Add front-end react app", "Add back-end nest app", and "Add types library", since those commits just generate stuff.**

This PR aims to get an initial Nx workspace started with both a React and NestJS application so that the CI and deployment can be tested.

Here is a screenshot of a Lighthouse test done after creating and serving a production build of the stack:

2020-08-23 00_52_08-DevTools - 127 0 0 1_8080_

Here is what you should expect to see once you have the stack running (prod or dev). If the "Welcome to rally-api!" message doesn't show up, it means the backend call failed, and you should check the console.
image


RGPosadas (approved):
# Deploying the Application
In the README, there were many ways to deploy the application in test/prod environments. I tested each one, and they all worked as expected passing the Lighthouse checks matching the screenshot above.

* [x] Serving locally in a dev environment
* [x] Serving locally in a prod environment
* [x] Serving the application as docker images

# Repo Structure
Cristian was kind enough to walk me through the repo structure of an nx workspace. What we currently have is pretty bare bones and is easy to understand (the lib/ folder already is a great improvement from minicap). 

A lot of files were generated which makes it kind of intimidating to go through, however they are defaults and are to be expected in such a workspace. I can already see the benefits comparing it to our repo structure from minicap. I am excited to work with nx!

# A Note About Dockerfiles
## Build Times
Cristian and I had a discussion about the long build times it takes for `npm i` to run and build inside the docker container. There are [ways to overcome this with docker volumes and perhaps other solutions](https://jdlm.info/articles/2016/03/06/lessons-building-node-app-docker.html), however these still need to be de-risked and tested since we don't know how they would interact with an nx workspace.

Time won't be as much of an issue if we're not constantly rebuilding the docker images for testing, and only rebuilding when ready to create a PR for example. However, building the setup repo already took a long time and there isn't much to begin with, so optimizing build times is definitely something that we want to have.

## Docker Best Practices
We can further enhance the Dockerfiles currently present in the repo by firmly going through [Docker Best Practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/) once we have more time. It can also be further optimized (like the point above) for JS applications, specifically one managed by an nx workspace!

# Conclusion
Looks good to me! Thanks for setting all this up @TheGreatMarkus !

r-saba (approved):
# Deploying the Application
No issues getting it running.

* [x] Serving locally in a dev environment
* [x] Serving locally in a prod environment
* [x] Serving the application as docker images


# Conclusion
Great work, the foundation is here.

RGPosadas:
GitHub is expecting a Travis check to succeed, however this branch does not contain a Travisfile, so I'm using my admin privileges to merge this PR since it has received 2 approvals.

PR #2, opened 2020-08-30
Setup Initial CI and CD

RGPosadas:
## Travisfile (CI)
You'll see that I activated caching in our Travisfile compared to minicap where we had it set to false. 
Setting cache `npm: true` caches `$HOME/.npm`, saving on build time.
Setting the pre-commit cache to be stored also makes it so that `pre-commit run -a` runs in 2 seconds instead of 60 like in minicap, since the python environment does not need to be recreated at every build.

## DockerHub (CI)
We currently have two DockerHub repos for the frontend and backend images of the application:
* [ritchellegmp/litterrally-frontend](https://hub.docker.com/repository/docker/ritchellegmp/litterrally-frontend)
* [ritchellegmp/litterrally-backend](https://hub.docker.com/repository/docker/ritchellegmp/litterrally-backend)

DockerHub AutoBuilds trigger a build when there is a merge/push to a branch and/or tag, according to defined AutoBuild rules.
There is currently one AutoBuild build rule for each image repository, and that is to build the `master` branch with the `latest tag`. As we go along with the semester, we'll have more AutoBuild rules once we have a more defined concept as to how we want things to be tagged and built.

## Kubernetes Manifests (CD)
You'll see that `k8s/kustomize/env/dev` and `k8s/kustomize/env/prod` paths contain the same kubernetes manifests, just with different names. It is possible that we will move from kustomize to helm for controlling manifests if the only difference between dev and prod are names of the manifests. For now, we will stick with kustomize since it has better handling of bigger discrepancies between environments as a way of future-proofing.

The files in `k8s/argocd-apps` enables ArgoCD to watch over our live applications, staying true to GitOps.

## Testing the PR
1. Download [kubectl CLI](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
2. Get a local cluster running, like [minikube](https://kubernetes.io/docs/setup/learning-environment/minikube/)
3. Then, simply follow `k8s/README.md`


cristian-aldea:
That's just a commit I forgot to push on my PR. It removes the .gitkeep files which aren't used anymore

k8s/README.md
```diff
@@ -0,0 +1,74 @@
...
+        `$ kubectl get pods -n argocd -l app.kubernetes.io/name=argocd-server -o name | cut -d'/' -f 2`
+
+     2. Port-forward the argocd service
+
+        `$ kubectl port-forward svc/argocd-server -n argocd 8080:443`
+
+     3. Go to `localhost:8080` and set
+        ```
+        user: admin
+        password: the pod name you get from step 1
```
cristian-aldea: Optional suggestion, but on the Github markdown preview the steps on that level shows as `a.`, `b.`, etc.
```suggestion
        password: the pod name you get from the first step
```

```suggestion
        password: The name of the Argo CD pod
```

k8s/argocd-apps/litterrally-dev.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: litterrally-dev
+  namespace: argocd
+spec:
+  project: default
+  source:
+    repoURL: https://github.com/RGPosadas/LitterRally.git
+    targetRevision: setup-ci-cd
```
cristian-aldea: This and a similar line from [k8s/argocd-apps/litterrally-prod.yaml](https://github.com/RGPosadas/LitterRally/pull/2/files#diff-d6daeb7faad82bb47441f74acacc6de4R10) will need to be updated before this is merged.

```suggestion
    targetRevision: master
```

k8s/README.md
```diff
@@ -0,0 +1,74 @@
...
+   Once logged in to ArgoCD, click `SYNC` on each application environment to sync the ArgoCD applications, which will finally deploy the LitterRally applications.
+
+5. Access the Applications
+
+   We currently don't have an ingress installed, so we have to port-forward the frontend and backend.
+
+   ```
+   $ export LITTERRALLY_POD=$(kubectl get pods -n litterrally-[dev|prod] | awk '{print $1}' | sed -n '2 p')
+   $ kubectl -n litterrally-[dev|prod] port-forward $LITTERRALLY_POD 9000:8080
+   $ kubectl -n litterrally-[dev|prod] port-forward $LITTERRALLY_POD 3333:3333
```
cristian-aldea: First I had some issues getting this to work since the actual port for the front-end for our current pod is 80, not 8080.

Second, I managed to get it to work with something simpler:
```suggestion
   $ export LITTERRALLY_POD=$(kubectl get pods -n litterrally-[dev|prod] | awk '{print $1}' | sed -n '2 p')
   $ kubectl -n litterrally-[dev|prod] port-forward $LITTERRALLY_POD 9000:80 3333
```

.travis.yml
```diff
@@ -0,0 +1,15 @@
+language: node_js
+node_js:
+  - '12.18.3'
+cache:
+  npm: true
```
cristian-aldea, line 5: Just a heads up I heard that enabling cache for node_modules might cause issues: 

https://stackoverflow.com/questions/42521884/should-i-have-travis-cache-node-modules-or-home-npm

This issue might have been fixed, we'll just need to keep an eye out for it.

Otherwise, enabling caching is gonna be an awesome change.

.travis.yml
```diff
@@ -0,0 +1,15 @@
...
+  directories:
+    - $HOME/.cache/pre-commit
+before_script:
+  - pyenv global 3.7.1
+  - pip3 install pre-commit
+script:
+  - pre-commit run -a
+# To uncomment when we have meaningful code
+# - npm install --log-level=error
+# - npm run lint
```
cristian-aldea, line 15: I'd say it wouldn't hurt to uncomment this now to make sure we don't forget to activate these checks. I actually caught a few linting errors that should be fixed before this is merged

cristian-aldea (dismissed):
I was able to get everything deployed without much trouble. Left a few comments with suggestions.

I have a small concern/reminder: I feel like having each app in separate pods and each with its own service and ingress is gonna be better in the long run. For now its okay, but we'll need to not forgot to create a ticket later imo.

k8s/README.md
RGPosadas: Whoops yeah dangers of copy paste, I initially had it hitting the service and decided to change to the pod, but forgot to update the port 👍

RGPosadas: Also didn't know the second way was a thing to begin with, TIL!

.travis.yml
RGPosadas: I actually also ran lint-fix and it did auto-fix some code, but there was one that had to be manually changed, but I was unsure of how to approach that due to my inexperience. Feel free to push code so that `npm run lint` works though!

.travis.yml
RGPosadas: I also saw this when setting the cache to true, but I also read that you can easily purge the Travis cache in case some inconsistencies start showing up between Travis builds and local testing! Perhaps we can add a warning line above this.

k8s/argocd-apps/litterrally-dev.yaml
RGPosadas: Thanks for the reminder!

k8s/README.md
RGPosadas: Oh interesting, on VSCode preview it doesn't show any numbers/letters at all. Will rewrite this a different way then

.travis.yml
RGPosadas: Ended up adding in a Troubleshooting part in the README regarding cache instead of an awkward comment

RGPosadas:
@TheGreatMarkus 
Thanks for catching those small inconsistencies!
Added a troubleshooting section to the README to make people aware of the possible issues with caching npm on Travis.
Feel free to push the fix to make the `npm run lint` pass in Travis!

cristian-aldea (approved):
Everything looks good to me!

RGPosadas:
Merging with my admin powers!

PR #3, opened 2020-09-02
Update prettier version and config

cristian-aldea:
Also formatted and linted the code.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -24,9 +24,7 @@
       {
         "enforceBuildableLibDependency": true,
         "allow": [],
-        "depConstraints": [
-          { "sourceTag": "*", "onlyDependOnLibsWithTags": ["*"] }
-        ]
+        "depConstraints": [{ "sourceTag": "*", "onlyDependOnLibsWithTags": ["*"] }]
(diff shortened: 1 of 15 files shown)


osahtout (dismissed):
Can confirm that the width has increase and prettier works

cindyslittleplanet (dismissed):
LGTM!
Prettier runs and a larger width can be seen as the code doesn't trail unnecessarily to the next line.

ghiobi (approved):
Great Müll type example from the lib and reformatting.

ghiobi:
Do you guys have a commit format to follow. Maybe something like https://github.com/commitizen/cz-cli could be followed? #suggestion

cristian-aldea:
> Do you guys have a commit format to follow. Maybe something like https://github.com/commitizen/cz-cli could be followed? #suggestion

We had a wiki which detailed our format for commits, but an automated system like the one you linked could be interesting, I'll look into it

osahtout (approved):
Still working as expected

cindyslittleplanet (approved):
Looking good!

PR #4, opened 2020-09-03
Change name to Mull

cristian-aldea:
All instances in text and files with "litter", "rally", "litter-rally" and "litterrally" were replaced with mull (Keeping the same case however)

If you have the time play around with this branch and test if everything still works. I tested as much as I could, but I might have missed something.


ys-lin (approved):
LGTM

leojrsilao (approved):
LGTM! 🚀 🚀

osahtout (approved):
I didn't see any "rally" or "litle" anywhere

PR #5, opened 2020-09-04
Update README.md and Travisfile

RGPosadas:
This PR aims to:
- Update Travis to run, at the very least, `npm run lint`
   - Please note that `npm install` is no longer needed as [Travis runs `npm ci` by default](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#npm-ci-support) due to the fact that we have a `package-lock.json`.
- Add a troubleshooting section that can be further expanded in the future
   - Please note that I also tried to add a ToC checker as a pre-commit, but the main 3 ones that I found had issues properly working as a pre-commit. Therefore, manually added ToC for README's through [github-markdown-toc ](https://github.com/ekalinin/github-markdown-toc)for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: node_js
 node_js:
   - '12.18.3'
+git:
+  depth: false
 cache:
   npm: true
   directories:
@@ -10,6 +12,4 @@ before_script:
   - pip3 install pre-commit
 script:
   - pre-commit run -a
-# To uncomment when we have meaningful code
-# - npm install --log-level=error
-# - npm run lint
+  - npm run affected:lint -- --base=origin/master
--- a/README.md
+++ b/README.md
@@ -1,5 +1,37 @@
 # Müll
 
+- [Team Members](#team-members)
+- [Development Guidelines](#development-guidelines)
+- [Setting up a development environment](#setting-up-a-development-environment)
+- [Running the application](#running-the-application)
+  - [Serving the application locally for development:](#serving-the-application-locally-for-development)
+  - [Building the application for production:](#building-the-application-for-production)
+  - [Serving the production built stack locally:](#serving-the-production-built-stack-locally)
+- [Building and running the application with Docker](#building-and-running-the-application-with-docker)
+  - [Front-end](#front-end)
+  - [Back-end](#back-end)
+- [Troubleshooting](#troubleshooting)
+  - [Inonsistent/Failing Travis Builds](#inonsistentfailing-travis-builds)
+
+# Team Members
+
+Müll is brought to you by members of `Team Bug bytes`.
+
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -10,6 +10,4 @@ before_script:
   - pip3 install pre-commit
 script:
   - pre-commit run -a
-# To uncomment when we have meaningful code
-# - npm install --log-level=error
-# - npm run lint
+  - npm run lint
```
cristian-aldea: Consider using the affected prefix, which will only run on projects and libs that were changed from HEAD master. `npm run lint` will only run linting on the ui project (if you don't specify a project it seems to take the first one)
```suggestion
  - npm run affected:lint
```

Or, if you really want to run linting for all projects, have a look here: https://nx.dev/angular/cli/run-many
Something like this should work:
1. Add a script in the package.json:
```json
"lint-all": "nx run-many --target=lint --all"
```
then you can write the follow in the travis file:
```suggestion
  - npm run lint-all
```

cristian-aldea (dismissed):
I'm not gonna block the PR with a "Request Changes" since I'll be out this afternoon, but please have a look at my comment, `nx lint` doesn't do what you would expect I'm afraid

.travis.yml
RGPosadas: Ended up using `nx run-many --all --target=lint` as you mentioned on Slack instead

RGPosadas: Oh derp I did not read this properly at all LOL, yeah that'll be better and shorter for devs as well. Will add it

cindyslittleplanet (dismissed):
LGTM!
The README.md is up to date and the Travis file has been updated.

.travis.yml
```diff
@@ -10,6 +10,4 @@ before_script:
   - pip3 install pre-commit
 script:
   - pre-commit run -a
-# To uncomment when we have meaningful code
-# - npm install --log-level=error
-# - npm run lint
+  - npm run lint-all
```
cristian-aldea: I think its a better idea to only lint/test/build the stuff that was changed in a given pr. Since Nx offered the option I thought it was pretty cool.

cristian-aldea (dismissed):
Everything looks good, just had a comment if you have the time.

RGPosadas:
FYI:
Disabled `--depth` for `git clone` in order for `origin/master` to be valid for our linting step. According to Travis, `--depth` is only useful if we have multiple git commands, which we don't, so this shouldn't conflict with anything in the future.
https://docs.travis-ci.com/user/customizing-the-build#git-clone-depth

cindyslittleplanet (approved):
Still LGTM!

cristian-aldea (approved):
Thanks for updating the docs!

Issue #10, opened 2020-09-20
TASK-2: Setup GraphQL


cristian-aldea:
**Task Description**

As a developer, I want GraphQL to be set up for communication between the front-end and back-end applications so that we can take advantage to GraphQL's powerful feature set when querying information.

**Acceptance Criteria**

- [ ] Have a working GraphQL setup
- [ ] Have functioning GraphQL calls 
- [ ] Have mock data to test

Issue #11, opened 2020-09-20
TASK-3: Setup TravisCI to lint, test and build projects when relevant


cristian-aldea:
**Task Description**

As a developer, I want the correct projects to be linted, tested and built depending on the context so that resources are utilized optimally.

**Extra Description**

The Nx CLI offers a powerful feature called affected, where Nx determines what projects in your workspace were affected by newly introduced changes so that only the relevant parts of the codebase are linted, tested and built. Nx does it by comparing the current HEAD with a BASE, both of which can be set manually. By default, the BASE is master, but this isn't always going to be accurate, since features branches will be merged into develop, not master. In such cases, develop should be the BASE, so that the correct diff can be calculated.

Hints:
* TravicCI lets you set environment variables to 
* https://docs.travis-ci.com/user/docker/
* https://nx.dev/angular/cli/affected

**Acceptance Criteria**
Modify the travis config yaml to do the following:
* [ ] If the current branch is master, lint, test and build all projects
* [ ] If the current branch is develop, only lint, test and build affected projects, and set BASE to master (default)
* [ ] If the current branch is a feature branch, only lint, test and build affected projects, and set BASE to develop (default)

cristian-aldea:
Removed the following AC from the ticket:
* [ ] Travis needs to trigger builds on DockerHub when appropriate (see above). Look if dockerhub can be triggered to build a project instead of listening to commits.

since it is covered by #50

PR #47, opened 2020-10-15
TASK-2: Setup GraphQL

bit172:
**Task Description**
This PR addresses #10.

**Note about GraphQL decorators and code coverage**
If you see ```/* istanbul ignore next */``` on any of the GraphQL classes, it is to ignore them from the code coverage since the decorators take in anonymous functions.  
We have tried different ways of fixing it, like [this](https://github.com/gstamac/ts-jest/blob/master/README.md#ignore-coverage-on-decorators). The creator of type-graphql even dismissed an issue about a fix. [LINK](https://github.com/MichalLytek/type-graphql/issues/514) 😥

**Testing Steps**
1. Make sure to resolve all your dependencies with `npm i`
1. Run the application using ```nx serve mull-api```.
1. If you see something like this in the logs: 
image 
It means that the GraphQL module has been loaded.
1. Open your browser and go to ```localhost:3333/graphql``` to enter the GraphiQL playground.
1. Test some queries and mutations. You should have the database open to check if your changes occur for mutations.

Additional information can be obtained by pressing on the DOCS and SCHEMA tabs or by watching the knowledge handover on [Slack](https://bug-bytes.slack.com/archives/CS5P4NL5Q/p1602637335001700).

 Example queries:
```graphql
{
  user(id:13) {
    id
    name
    description
    password
  }
}
```
```graphql
{
  user(id:12) {
    id
    name
    description
    password
    friends {
      name
    }
  }
}
```

Example mutations:
```graphql
mutation {
  createUser(createUserInput: {
    name:"George"
    email:"ge@orge.ca"
    password:"password"
    age:"22"
  }) {
    id
  }
}
```
```graphql
mutation {
  updateUser(updateUserInput: {
    id: 14
    age: "23"
  }) {
    id
  }
}
```


r-saba (changes requested):
As @TheGreatMarkus pointed out. We don't need both `user.type.ts` and `user.entity.ts`. You can just move the decorators into one file and use that as your user type. I tested it out locally and it works fine.

You may have to update your mock data though, keep me updated.

*I have not finished reviewing yet, just pointing this out so you can get working on it early*

apps/mull-api/src/app/user/user.service.ts
```diff
@@ -0,0 +1,48 @@
...
+    return friends;
+  }
+
+  async create(userInput: CreateUserInput): Promise<User> {
+    const { age, name, email, password } = userInput;
+    const newUser = this.userRepository.create();
+    newUser.name = name;
+    newUser.email = email;
+    newUser.password = password;
+    newUser.age = age;
```
r-saba: To make this easier to read you can pass in the entity properties as an object. Like below. Note that since the property name matches the value name we don't need `email: email` we can just write `email`

`const newUser = this.userRepository.create({
      name,
      email,
      password,
      age
    });`

r-saba (changes requested):
Looking really good! All working well. Just some code quality comments.

1. Can we get rid of the code duplication due to `user.type.ts` and `user.entity.ts` being the same class?
2. Can we move the tests/mock data into another directory?

.travis.yml
```diff
@@ -17,7 +17,7 @@ addons:
       secure: $SONAR_TOKEN
 script:
   - pre-commit run -a
-  - npm run affected:lint -- --base=origin/master
-  - npm run affected:test -- --base=origin/master --codeCoverage
-  - npm run affected:e2e -- --base=origin/master
+  - npm run affected:lint -- --base=origin/develop
+  - npm run affected:test -- --base=origin/develop --codeCoverage
+  - npm run affected:e2e -- --base=origin/develop
```
cristian-aldea, line 22: Just a note, this will need to be addressed by #11 before our first release, because if it gets merged to master/main then the comparison won't make sense anymore haha

apps/mull-api/src/app/user/user.service.ts
bit172: Good Idea @r-saba! That looks a lot cleaner. 🧹
As for testing, Nx generates test files next to the source files and I think we should just stick to what it provides us. 🗃

r-saba (approved):
LGTM! Thanks for addressing the comments above. 👏
**Acceptance Criteria**

- [X] Have a working GraphQL setup
- [X] Have functioning GraphQL calls 
- [X] Have mock data to test

cristian-aldea (approved):
Everything looks good. Thanks for removing the UserType class you had to make things a bit less redundant @bit172!

**Acceptance Criteria**

- [x] Have a working GraphQL setup
- [x] Have functioning GraphQL calls 
- [x] Have mock data for testing

cindyslittleplanet (approved):
LGTM! Thank you for setting up GraphQL! :)

**Acceptance Criteria**
- [X] Have a working GraphQL setup
- [X] Have functioning GraphQL calls 
- [X] Have mock data to test

apps/mull-api/src/app/user/user.service.ts
ghiobi: Do you guys, encrypt the password before saving it to the database?

bit172: Not yet, but that would be done for user registration using a package like bcrypt. 🚀

ghiobi:
```
  create(userInput: CreateUserInput): Promise<User> {
    return this.userRepository.save(userInput);
  }
```
should work too.

PR #52, opened 2020-10-19
FIX: mull-api Entities

cristian-aldea:
This PR aims to fix the circular dependency build issues that @r-saba encountered when working on TypeORM.

Also, when working on the class diagram, a few issues came up as to how intuitive the entities are to use, this PR aims to fix that too.

See commit messages for details.

Here is the resulting ER diagram from these changes
final


bit172 (dismissed):
LGTM! We will need this merged before continuing with US1.1.

apps/mull-api/src/schema.gql
```diff
@@ -3,12 +3,17 @@
...
   name: String!
   password: String!
 }
 
+"""
+A date-time string at UTC, such as 2019-12-03T09:54:33Z, compliant with the date-time format.
+"""
+scalar DateTime
+
 type Mutation {
```
ghiobi, line 17: Probably should be ignored cuz it's generated. Create a gitignore in apps/mull-api so we modularize our gitignore files and add.
`/src/schema.gql`

workspace.json
```diff
@@ -120,7 +120,8 @@
             "outputPath": "dist/apps/mull-api",
             "main": "apps/mull-api/src/main.ts",
             "tsConfig": "apps/mull-api/tsconfig.app.json",
-            "assets": ["apps/mull-api/src/assets"]
+            "assets": ["apps/mull-api/src/assets"],
+            "showCircularDependencies": false
```
ghiobi, line 124: Disabling the detection of circular dependencies is probably not a good idea. It actually very useful and will help you from running into circular depend bugs due to the nature of how Node modules work. It's your worst nightmare.

tsconfig.base.json
```diff
@@ -9,7 +9,7 @@
     "experimentalDecorators": true,
     "importHelpers": true,
     "target": "es2015",
-    "module": "esnext",
+    "module": "commonjs",
```
ghiobi: going back to the old days?

apps/mull-api/src/schema.gql
cristian-aldea: I think it might be a good idea to leave it in version control, to make sure everyone has it and to keep track of changes to our schema, kinda like how package-lock.json is in version control usually.

tsconfig.base.json
cristian-aldea: Using the CommonJS module system avoids the building issue Ragith and I was having with TypeORM. These issues were caused by circular dependencies I created by refactoring the entities

workspace.json
cristian-aldea: I would agree with you, but with something like typeorm, creating circular dependencies is very normal, for example with something like a OneToMany relationship, which requires two classes to import each other. See: 
- nrwl/nx#2975
- nrwl/nx#1667

These warnings aren't helpful in this case, but maybe there's a more granular way to ignore these warning so that other parts still get warnings

ghiobi: Yes, it is very normal to have circular dependencies in es modules to a certain degree, but definitely one can run into issues if they aren't cautious. In fact, TypeOrm allows us to use strings for these types of issues as it was pushed below.

https://github.com/typeorm/typeorm/pull/4195

So the question is, would you want to disable circular dependency detection and run into potential issues in the future. Or use something TypeOrm already fixed?

cristian-aldea: Using strings in the annotations won't fix the circular dependency warnings because of the changes I did to the models
I made a branch on which I turned back the warnings on for circular dependencies and reverted back to using strings:`FIX/mull-api-entities-test`, on which you can run the backend

tsconfig.base.json
ghiobi: I see, alright let's make this problem of the future, and implement some features. this is no biggie.

workspace.json
ghiobi: Let's make this problem of the future, and implement some features. This is no biggie as long as the team knows what circular imports are.

leojrsilao (dismissed):
LGTM!! Good work on the fixes for the circular dependency issues of TypeORM 🚀 🚀

tsconfig.base.json
r-saba: Unfortunate but I do think this is the best solution for this project. When I moved over to using strings like @ghiobi mentioned I no longer had the issue of accessing a class before it was initialized. The refactoring in this PR probably introduced tighter coupling between entities but makes them more intuitive to work with.

Another benefit is we can use types directly in TypeORM decorators now, making our code easier to read and traverse.

r-saba (dismissed):
LGTM! Good work @TheGreatMarkus

tsconfig.base.json
cristian-aldea: I will be moving this declaration of module system into the mull-api's tsconfig, to avoid affecting other apps.

cristian-aldea:
Sorry for dismissing all of your reviews @r-saba @bit172 @leojrsilao, but the adjustment @ghiobi mentioned was important. I made it so ONLY mull-api uses the older commonjs module system, so as to minimize the impact of this PR on other apps.

Feel free to give a quick LGTM when you can

apps/mull-api/src/schema.gql
cristian-aldea: We can add it to gitignore in a separate PR if need be, since these changes are urgent, sorry laurendy

bit172 (approved):

r-saba (approved):

PR #71, opened 2020-11-10
FIX: Remove unused proxy conf file

cristian-aldea:
This file created a proxy for the backend, allowing us to access the backend through localhost:4200/api or localhost:4200/graphql. It was generated when the project was created but is no longer used

PR #74, opened 2020-11-11
HOTFIX: Remove proxyConfig option for mull-ui serve

cristian-aldea:
I'm very sorry everyone for this oversight, but the proxy config file deleted in #71 was referenced in the workspace.json, and serving the front-end doesn't work unless that reference is removed.

Error:
image

### DIFF
--- a/workspace.json
+++ b/workspace.json
@@ -56,8 +56,7 @@
         "serve": {
           "builder": "@nrwl/web:dev-server",
           "options": {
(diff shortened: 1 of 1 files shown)


ys-lin (approved):
Everything seems to be running properly 👍

r-saba (approved):
Able to serve the UI. LGTM!

Issue #28, opened 2020-09-27
TASK-27: Upload Event Page Photo


RGPosadas:
**User Story Description**

As a host, I want to upload a photo to my event so that participants are able to view the littered area.

**User Acceptance Flow**

1. While creating an event, user has to upload a photo as their event image
1. When user submits the event form, the uploaded image becomes the event preview image

**Acceptance Criteria**

- [ ] User acceptance flow is respected
- [ ] Documentation present where needed
- [ ] Unit and UI tests are implemented for every component and function
- [ ] There must exist an automated Cypress acceptance test
- [ ] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=2000%3A222910)

PR #59, opened 2020-11-04
US-1.1: Create Events

JimLau49:
**Task Description**
This PR addresses #27 

Create events is now working (without image and location uploading) with validation for supported inputs. After entering valid form data, users will be able to send their events to our database.

**To run the E2E tests**
1. Make sure to run `npm i ` on the project to pickup the new dependencies 
1. Run the server `nx serve mull-api`
1. In the root directory run  `nx e2e mull-ui-e2e --watch` ( Cypress takes some time to open up)
**Note:** for `nx e2e mull-ui-e2e --watch` , the `--watch` option keeps Cypress open to watch for code changes instead of immediately running the tests and closing.
1. When the cypress window opens up select `create-event.spec.ts` and watch the tests go 🚀 
image

Issue #27, opened 2020-09-27
US-1.1: Create Events


RGPosadas:
**User Story Description**

As a user, I want to create and edit events so that I can coordinate cleanup activities with other users.

**User Acceptance Flow**

1. User presses "Create Event" button on the navigation bar
1. User fills in an event form (without uploading photo and location)
   a. If user submits invalid data, then error messages show up
1. User submits event form
1. User gets a confirmation that the event was created successfully.

**Acceptance Criteria**

- [ ] User acceptance flow is respected
- [ ] Documentation present for every new component and function
- [ ] Unit and UI tests are implemented for every component and function
- [ ] There must exist an automated Cypress acceptance test in GIF form
- [ ] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/?node-id=534%3A6528)

PR #77, opened 2020-11-12
US-1.2: upload event page photo

r-saba:
**Task Description**
#28  

File uploading using the Create Event form is now working. The image is a required field and is validated when the form is submitted. When the form is successfully submitted the image is stored in the mull-api server and its filename is `id+extension` where id comes from the row id in the Media table.

The fileupload field was refactored into its own CustomFileUpload component.


libs/ui-lib/src/lib/custom-file-upload/custom-file-upload.tsx
```diff
@@ -0,0 +1,42 @@
...
+      {hasErrors ? <span className="error-message">{errorMessage}</span> : null}
+      <input
+        className="event-image-upload"
+        id="imageFile"
+        type="file"
+        onChange={handleFileUpload}
+        accept="image/*"
+        data-testid="file"
+      />
+    </>
```
bit172: Would this be better if it were in a div instead of a fragment? In case we want to style the entire component or make use of a container for this component.

apps/mull-ui/src/app/pages/create-event/create-event.tsx
```diff
@@ -153,19 +178,13 @@ const CreateEventPage = ({ history }) => {
...
-            type="file"
-            onChange={handleFileUpload}
-          />
+          <CustomFileUpload
+            imageURL={imageURLFile}
+            hasErrors={formik.touched.imageFile && !!formik.errors.imageFile}
+            errorMessage={formik.errors.imageFile}
+            handleFileUpload={handleFileUpload}
+            uploadIcon={<UploadIcon />}
+          ></CustomFileUpload>
```
bit172:
```suggestion
          <CustomFileUpload
            imageURL={imageURLFile}
            hasErrors={formik.touched.imageFile && !!formik.errors.imageFile}
            errorMessage={formik.errors.imageFile}
            handleFileUpload={handleFileUpload}
            uploadIcon={<UploadIcon />}
          />
```
This component doesn't need children. So this syntax is cleaner.

apps/mull-ui/src/app/pages/create-event/create-event.tsx
```diff
@@ -111,19 +129,26 @@ const CreateEventPage = ({ history }) => {
...
     }),
 
-    onSubmit: (values) => {
+    onSubmit: async (values) => {
       notifySubmissionToast();
       if (!values.endDate) values.endDate = cloneDeep(values.startDate);
       addTimeToDate(values.startTime, values.startDate);
       addTimeToDate(values.endTime, values.endDate);
+      let media = await uploadFile({ variables: { file: file } });
+      let imageMedia: MediaType = {
```
bit172:
```suggestion
      const media = await uploadFile({ variables: { file: file } });
      const imageMedia: MediaType = {
```
Since these are never reassigned.

apps/mull-ui/src/main.tsx
```diff
@@ -23,7 +25,9 @@ if ('serviceWorker' in navigator) {
 }
 
 const client = new ApolloClient({
-  uri: environment.backendUrl,
+  link: createUploadLink({
+    uri: environment.backendUrl,
+  }),
```
bit172, line 30: This is just a comment. Just want to know what this is for.

apps/mull-ui/src/app/pages/create-event/create-event.tsx
```diff
@@ -21,6 +21,11 @@ export interface CreateEventProps {
   history: History;
 }
 
+export interface MediaType {
+  id: number;
+  mediaType: string;
+}
+
```
bit172: Can this be moved into our types lib?

libs/ui-lib/src/lib/custom-file-upload/custom-file-upload.tsx
```diff
@@ -0,0 +1,42 @@
...
+        {imageURL ? (
+          <img src={imageURL} style={{ width: '50%', height: '50%' }} alt="Event" />
+        ) : (
+          uploadIcon
+        )}
+      </label>
+      {hasErrors ? <span className="error-message">{errorMessage}</span> : null}
+      <input
+        className="event-image-upload"
+        id="imageFile"
```
bit172: Would probably be a good idea to make this id a prop along with the label in case there are multiple CustomFileUpload components on the same page.

apps/mull-ui/src/app/pages/create-event/create-event.tsx
```diff
@@ -111,19 +129,26 @@ const CreateEventPage = ({ history }) => {
...
+      imageFile: Yup.mixed().required('Image File is required.'),
     }),
 
-    onSubmit: (values) => {
+    onSubmit: async (values) => {
       notifySubmissionToast();
       if (!values.endDate) values.endDate = cloneDeep(values.startDate);
       addTimeToDate(values.startTime, values.startDate);
       addTimeToDate(values.endTime, values.endDate);
+      let media = await uploadFile({ variables: { file: file } });
```
bit172: There should be a check for if this promise is resolved or rejected. I suspect this is the cause of my submission without the backend problem.

bit172 (changes requested):
Looks good! Just a few things to fix.

Also, when I try submitting without the backend running, I do not get an error on the client. It is stuck and nothing happens. Do you know what might be causing this issue?
image

apps/mull-ui/src/app/pages/create-event/create-event.tsx
bit172: Consider using a try-catch block or using .then() .catch() instead of await.

apps/mull-ui/src/main.tsx
r-saba: By default apollo V3 wraps the uri with HttpLink which is a terminating link that fetches GraphQL results from a GraphQL endpoint over an http connection. The createUploadLink explicitly defined here allows us to use File instances within query or mutation variables. Without this we cannot send files to the end point using Apollo. 
To learn more - https://www.npmjs.com/package/apollo-upload-client

apps/mull-ui/src/app/pages/create-event/create-event.tsx
```diff
@@ -111,19 +129,26 @@ const CreateEventPage = ({ history }) => {
         .required('Event Description is required.')
         .max(5000, 'Event Description must be under 5000 characters.'),
       location: Yup.string().required('Event Location is required.'),
+      imageFile: Yup.mixed().required('Image File is required.'),
```
JimLau49: This could be 'Image is required' 'Image file' doesn't sound as natural

apps/mull-ui/src/app/pages/create-event/create-event.tsx
TheGreatMarkus: Yes please move this to our types library and have our Media entity implement the interface

apps/mull-ui/src/app/pages/create-event/create-event.spec.tsx
```diff
@@ -37,7 +37,7 @@ describe('CreateEvent', () => {
       fireEvent.click(submitButton);
     });
     const error = utils.container.querySelector('span[class="error-message"]');
-    expect(error).toHaveTextContent('Start Time is required');
+    expect(error).toHaveTextContent('Image File is required');
```
JimLau49: If you made the change above then, this should also be changed to 'Image is required'

libs/ui-lib/src/lib/custom-file-upload/custom-file-upload.tsx
```diff
@@ -0,0 +1,42 @@
+import React, { ChangeEvent, ReactNode } from 'react';
+
+import './custom-file-upload.scss';
+
+export interface CustomFileUploadProps {
+  hasErrors: boolean;
+  errorMessage: string;
+  imageURL: string;
+  handleFileUpload: (event: ChangeEvent<HTMLInputElement>) => void;
+  uploadIcon: ReactNode;
```
cristian-aldea: Does the icon for this component really need to be passed? I feel like it could just be part of the component, either by including the svg in the UI lib, or moving this component to the mull-ui app

apps/mull-ui/src/app/pages/create-event/create-event.tsx
```diff
@@ -39,15 +53,18 @@ const CreateEventPage = ({ history }) => {
   const toastId = useRef(null);
   // GraphQL mutation hook to create events
   const [createEvent] = useMutation(CREATE_EVENT);
+  const [uploadFile] = useMutation(UPLOAD_PHOTO);
   // Image of Event
-  const [imageFile, setImageFile] = useState(null);
-
+  const [imageURLFile, setImageURLFile] = useState(null);
+  const [file, setFile] = useState<File>(null);
```
JimLau49: We could add a line of documentation explaining what each hook is doing, from what i understand, the need for 2 hooks is so that we can keep the image preview on the page. So adding an extra line would help other dev's down the road who are working on this component

apps/mull-ui-e2e/src/integration/create-event.spec.ts
```diff
@@ -61,17 +67,20 @@ frameSizes.forEach((frame) => {
...
         'Event Description is required.'
       );
-      cy.get(':nth-child(9) > .error-message').should('have.text', 'Event Location is required.');
+      cy.get(':nth-child(10) > .error-message').should('have.text', 'Event Location is required.');
     });
 
     it('should show a successful submission message', () => {
+      cy.get('[data-testid="file"]').attachFile('../fixtures/zoro.jpg');
+
       cy.get(':nth-child(5) > [data-testid=custom-time-input]').type('11:20');
```
JimLau49, line 85: Unnecessary white space here

apps/mull-ui-e2e/src/integration/create-event.spec.ts
```diff
@@ -61,17 +67,20 @@ frameSizes.forEach((frame) => {
...
         'have.text',
         'Event Description is required.'
       );
-      cy.get(':nth-child(9) > .error-message').should('have.text', 'Event Location is required.');
+      cy.get(':nth-child(10) > .error-message').should('have.text', 'Event Location is required.');
     });
 
     it('should show a successful submission message', () => {
+      cy.get('[data-testid="file"]').attachFile('../fixtures/zoro.jpg');
+
```
JimLau49, line 83: Unnecessary white space here

apps/mull-api/src/app/event/inputs/event.input.ts
```diff
@@ -45,6 +50,18 @@ export class UpdateEventInput implements Partial<Event> {
...
 }
+
+@InputType()
+export class MediaInput implements Partial<Media> {
+  @Field()
+  id: number;
+
+  @Field()
+  mediaType: string;
+}
```
cristian-aldea: Shouldn't this be in the media module under a file like app/media/input/media.input.ts?

JimLau49 (changes requested):
Looking good so far, however im getting an error when trying to submit **with** the backend running and im getting this error message in mull-api
`RangeError: Maximum call stack size exceeded
    at ReadStream.<anonymous> (internal/fs/streams.js:131:3)
    at ReadStream.deprecated [as open] (internal/util.js:89:15)
  `
image
Same error as @bit172 but with the backend running too.

apps/mull-ui/src/app/pages/create-event/create-event.tsx
```diff
@@ -30,6 +35,15 @@ const CREATE_EVENT = gql`
...
+    uploadFile(file: $file) {
+      id
+      mediaType
+    }
+  }
+`;
+
 /**
  * This component renders the create event page
  * @param {History} history
```
cristian-aldea, line 45: Can you replace 
```tsx
const CreateEventPage = ({ history }) => {
```

with 

```tsx
const CreateEventPage = ({ history }: CreateEventProps) => {
```

Sorry that I didn't catch this in the previous PR

apps/mull-ui/src/app/pages/create-event/create-event.tsx
```diff
@@ -21,6 +21,11 @@ export interface CreateEventProps {
   history: History;
```
cristian-aldea:
```suggestion
  history: IHistory;
```

and import the right History interface

```ts
import { History as IHistory } from 'history';
```

libs/ui-lib/src/lib/custom-file-upload/custom-file-upload.tsx
```diff
@@ -0,0 +1,42 @@
...
+}: CustomFileUploadProps) => {
+  return (
+    <>
+      <label htmlFor="imageFile" className="custom-file-upload event-input-border">
+        {imageURL ? (
+          <img src={imageURL} style={{ width: '50%', height: '50%' }} alt="Event" />
+        ) : (
+          uploadIcon
+        )}
+      </label>
```
cristian-aldea:
```suggestion
      <label
        htmlFor="imageFile"
        style={{ boxSizing: 'border-box' }}
        className="custom-file-upload"
      >
        {imageURL ? (
          <img
            src={imageURL}
            style={{ width: '100%', display: 'block', borderRadius: '6.5px' }}
            alt="Event"
          />
        ) : (
          <div
            style={{
              margin: '1.8em 5em',
              boxSizing: 'border-box',
              width: 'auto',
            }}
          >
            {uploadIcon}
          </div>
        )}
      </label>
```

this is to make the image upload look good. This was brought up in #59 and was supposed to be fixed in this PR.

Please remove the inline styling in my suggestion and place them in appropriate classes.

libs/ui-lib/src/lib/custom-file-upload/custom-file-upload.tsx
```diff
@@ -0,0 +1,42 @@
+import React, { ChangeEvent, ReactNode } from 'react';
```
cristian-aldea: (sorry i couldn't leave this comment in app.scss)

Please remove
```scss
div {
  width: 100%;
}
```
from app.scss and fix the styling.

cristian-aldea: Please move styling relevent  to the custom-file-upload.tsx component from `create-event.scss` to `libs/ui-lib/src/lib/custom-file-upload/custom-file-upload.scss`

cristian-aldea (changes requested):
Great work! I can confirm that uploading a picture works when filling the form, and the Media is successfully created in the database and the image successfully saved on the (currently local) media server).

I had a few nitpicks about styling since this ticket was meant to address some of the styling issues I had with #59, but those issues are still present here.

Also had a few issues with how interfaces are added in this PR. Please see my comments below

apps/mull-ui/src/app/pages/create-event/create-event.tsx
cristian-aldea: Why do we need two hooks

leojrsilao: Added a new hook because if I set the state for the same hook, then the preview of the image becomes a question mark image.

apps/mull-ui/src/app/pages/create-event/create-event.tsx
r-saba: Really nice catch and recommendation!! Thanks

r-saba:
All the comments have been addressed. Seems like I've accidentally combined a few of my semantic commits when rebasing.

I decided to rewrite all the selectors used in the create-event e2e tests. Every time we added a new element I would have to fix **every** single test, even those unrelated to this pr. It was very annoying, they're robust now. Even if an element is added or removed it won't break the existing e2e tests.
I also increased the timeout for the success toast to 5 seconds because 1 second was too short and kept failing the build. I'm still not sure why the build is failing though.

@JimLau49 I also pushed a fix for the issue you were having, more info on the issue [here](https://github.com/apollographql/apollo-server/issues/3508)

Thanks for the great feedback everyone! Let me know if you find anything else.

apps/mull-ui/src/app/pages/create-event/create-event.tsx
cristian-aldea: No need for the alias for the interface btw, just  `import { History } from 'history';` works fine

cristian-aldea:
I'll have a look again when I'm done with my c++ assignment

JimLau49:
Will have a look again after work tmr

.pre-commit-config.yaml
```diff
@@ -5,7 +5,9 @@ repos:
       - id: detect-private-key
       - id: check-yaml
       - id: end-of-file-fixer
+        exclude: 'package-lock.json'
   - repo: https://github.com/prettier/prettier
     rev: 2.1.1
     hooks:
       - id: prettier
+        exclude: 'package-lock.json'
```
cristian-aldea, line 13: Just wanted clarification on this, since you talked about your issues during our scrums.

Is this a fix you have the intention of merging to develop? It's not a big deal to have these exclusions, but its really unfortunate that other solutions didn't work

apps/mull-ui/src/app/pages/create-event/date-calendar/date-calendar.scss
```diff
@@ -1,6 +1,7 @@
 @import 'variables';
 
 .date-calendar-container {
+  width: 100%;
```
cristian-aldea, line 4: I'd suggest to remove this line you added
```suggestion
```
And add the following to `apps/mull-ui/src/app/pages/create-event/create-event.scss`
```scss
.date-calendar-container,
.custom-text-input-container {
  width: 100%;
}
```
Sinec both the calendar AND the text input need to have their width set to 100%.

apps/mull-api/src/app/media/media.resolver.ts
```diff
@@ -2,24 +2,31 @@ import { Resolver, Mutation, Args } from '@nestjs/graphql';
...
+        this.mediaService.updateFilename(file.filename, media.id, media.mediaType);
+        const newMedia = new Media(media.mediaType);
+        newMedia.id = media.id;
+        return newMedia;
       })
-      .catch(() => false);
+      .catch((e) => {
+        console.log(e);
+        return null;
+      });
```
cristian-aldea: I don't know how returning exceptions works with GraphQL, but shouldn't we return an error instead of `null`?
Also, I think `console.error` would be more relevant to use here

apps/mull-ui-e2e/src/integration/create-event.spec.ts
```diff
@@ -1,4 +1,5 @@
 /// <reference types="Cypress" />
```
cristian-aldea, line 1: Thanks for fixing this e2e test! Doing nth child queries really seemed weird to me

apps/mull-ui/.eslintrc
```diff
@@ -140,6 +140,7 @@
     ],
     "no-with": "warn",
     "no-whitespace-before-property": "warn",
+    "no-var": "off",
```
cristian-aldea: You can move this rule to the root `.eslintrc`, which will have it apply to our whole workspace.

apps/mull-ui/src/app/app.scss
```diff
@@ -1,9 +1,5 @@
 @import 'variables';
 
-div {
-  width: 100%;
-}
-
```
cristian-aldea: Thanks for removing this, it caused a bit of trouble for me in my ticket :x

apps/mull-ui/src/app/pages/create-event/create-event.tsx
cristian-aldea: Yeah, I'd really like a quick comment to explaining this, since it might be confusing

apps/mull-ui/src/app/pages/create-event/create-event.tsx
```diff
@@ -33,24 +29,36 @@ const CREATE_EVENT = gql`
...
   // Reference Id for the toast
   const toastId = useRef(null);
   // GraphQL mutation hook to create events
   const [createEvent] = useMutation(CREATE_EVENT);
+  const [uploadFile] = useMutation(UPLOAD_PHOTO);
   // Image of Event
-  const [imageFile, setImageFile] = useState(null);
-
+  const [imageURLFile, setImageURLFile] = useState(null);
+  const [file, setFile] = useState<File>(null);
```
cristian-aldea: If possible, add type annotation to hook to get that sweet autocomplete. I'm not sure what the mutation hooks need as type, so my suggestion might be wrong:
```ts
  const [createEvent] = useMutation<IEvent>(CREATE_EVENT);
  const [uploadFile] = useMutation<Upload>(UPLOAD_PHOTO);
  // Image of Event
  const [imageURLFile, setImageURLFile] = useState<string>(null);
  const [file, setFile] = useState<File>(null);
```

apps/mull-ui/src/app/pages/create-event/create-event.tsx
```diff
@@ -114,19 +123,32 @@ const CreateEventPage = ({ history }) => {
...
+      } catch (err) {
+        updateSubmissionToast(toast.TYPE.ERROR, 'Fatal Error: Event Not Created');
+        console.error(err);
+        return;
+      }
+      const imageMedia: IMedia = {
+        id: media.data.uploadFile.id,
+        mediaType: media.data.uploadFile.mediaType,
+      };
       const payload = {
```
cristian-aldea: You can set the type of this to:
```ts
const payload: Partial<IEvent> = {
```
for IEvent, you can have a look at my example [here](https://github.com/RGPosadas/Mull/pull/80/files#diff-175bb5751a40c4aac4cf8ebea1c5678bb9c7a2f6101bfe447d88d5ade6915135R19-R30).

libs/ui-lib/src/index.ts
```diff
@@ -1,3 +1,4 @@
+export * from './lib/custom-file-upload/custom-file-upload';
```
cristian-aldea: Please move this to mull-ui's component folder. As we discussed, ui-lib will be removed around the end of the sprint

libs/ui-lib/src/lib/custom-file-upload/custom-file-upload.scss
```diff
@@ -0,0 +1,50 @@
...
+  display: 'block';
+  border-radius: '6.5px';
+}
+
+.custom-file-upload {
+  display: block;
+  box-sizing: border-box;
+  cursor: pointer;
+  border-radius: 8px;
+  border: 1.5px $primary-color solid;
```
cristian-aldea: Partial pixel amounts sometimes leave whitespace. If you do this suggestion:
```suggestion
  border: 1px $primary-color solid;
```

And set:
```scss
.custom-file-upload-image {
  ...
  border-radius: 7px; // instead of 6.5px
}
```
You get this:
image

Instead of this:
image

libs/ui-lib/src/lib/custom-file-upload/custom-file-upload.tsx
```diff
@@ -0,0 +1,46 @@
...
+  hasErrors,
+  errorMessage,
+  handleFileUpload,
+  fieldName,
+}: CustomFileUploadProps) => {
+  return (
+    <div className="custom-file-upload-container">
+      <label htmlFor={fieldName} className="custom-file-upload">
+        {imageURL ? (
+          <img className="custom-file-upload-image" src={imageURL} alt="Event" />
```
cristian-aldea: @r-saba Brought up a good point that an img in a label doesn't make much sense. Any plans on refactoring this in this ticket? Not having an img in the label might also solve the styling issues I comment on above.

cristian-aldea (changes requested):
Everything looks awesome! Had some small comments. Will approve once those are addressed

apps/mull-ui-e2e/src/integration/create-event.spec.ts
JimLau49: Agreed

.pre-commit-config.yaml
r-saba: I spent some time with @RGPosadas and @bit172  looking into resolving the issue without finding a solution yet. I do plan to merge this into develop, we can track this as a bug.

libs/ui-lib/src/lib/custom-text-input/custom-text-input.spec.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { ChangeEvent } from 'react';
-import { fireEvent, render } from '@testing-library/react';
```
JimLau49: Thanks for removing this unused import

JimLau49 (dismissed):
- [x] User acceptance flow is respected
- [x] Documentation present for every new component and function
- [x] Must match [Mockup](create-mockup-link)

LGTM, thanks for making my suggested changes and also a special thanks to @r-saba for figuring out the issue with the `Maximum call stack size exceeded at ReadStream.` error.

bit172 (dismissed):
- [x] User acceptance flow is respected
- [x] Documentation present for every new component and function
- [x] Must match [Mockup](create-mockup-link)

LGTM! All my comments have been addressed. Good work!

libs/ui-lib/src/lib/custom-file-upload/custom-file-upload.scss
r-saba: Actually I noticed in the mockup there is no border when the image is uploaded. It also looks cleaner without it, so I remove the border when an image has been uploaded.

libs/ui-lib/src/lib/custom-file-upload/custom-file-upload.tsx
r-saba: Thanks for bringing that up again. I actually think the method @bit172 came up with here is ingenious and easier to read than what I was proposing. The solution I used in another project uses the useRef hook and is not as intuitive to follow. Doing it with pure HTML is much better and since `label` isn't some special element which requires only text, this works perfectly.

The styling issue isn't related to the `label` element.

apps/mull-ui/src/app/pages/create-event/date-calendar/date-calendar.scss
r-saba: I added `width 100%` to each class separately. It made more sense since `.custom-text-input-container` already exists in the file and has another style applied to it as well.

cristian-aldea: Aight sounds good

apps/mull-ui/src/app/pages/create-event/create-event.tsx
r-saba: That sweet sweet auto complete

apps/mull-ui/src/app/pages/create-event/create-event.tsx
r-saba: Agreed

JimLau49 (dismissed):

.pre-commit-config.yaml
TheGreatMarkus: Yes please create a bug before merging this PR @r-saba

apps/mull-ui/src/app/components/custom-file-upload/custom-file-upload.scss
```diff
@@ -0,0 +1,54 @@
...
+
+.imageURL {
+  margin: 1.8em 5em;
+  box-sizing: border-box;
+  width: auto;
+}
+.imageURL-active {
+  width: '100%';
+  display: 'block';
+  border-radius: '6.5px';
```
TheGreatMarkus, line 11:
```suggestion
  border-radius: 6.5px;
```

apps/mull-ui/src/app/pages/create-event/create-event.tsx
TheGreatMarkus: Saw you added comments, resolving this

apps/mull-ui/src/app/pages/create-event/date-calendar/date-calendar.scss
TheGreatMarkus: I see you added the styling to each class in create.event.scss. However now this line you added is dead code since its applied twice in two places. Please remove it.

libs/ui-lib/src/lib/custom-file-upload/custom-file-upload.scss
cristian-aldea: alright sounds good

apps/mull-ui/src/app/components/custom-file-upload/custom-file-upload.scss
cristian-aldea:
```suggestion
  border-radius: 6.5px;
```

apps/mull-ui/src/app/components/custom-file-upload/custom-file-upload.scss
```diff
@@ -0,0 +1,54 @@
+@import 'variables';
+
+.imageURL {
+  margin: 1.8em 5em;
```
cristian-aldea, line 4:
```suggestion
  margin: 1.8rem 5rem;
```

Please also remove other unnecessary uses of `em` in `create-event.scss`.

Or even better, remove all bad instances of `em` in our code. you can use the regex search `\dem` to find them

apps/mull-ui/src/app/pages/create-event/create-event.tsx
```diff
@@ -44,16 +48,19 @@ const CreateEventPage = ({ history }: CreateEventProps) => {
...
   const toastId = useRef(null);
   // GraphQL mutation hook to create events
-  const [createEvent] = useMutation(CREATE_EVENT);
-  // Image of Event
-  const [imageFile, setImageFile] = useState(null);
-
+  const [createEvent] = useMutation<IEvent>(CREATE_EVENT);
+  const [uploadFile] = useMutation<{ uploadFile: IMedia }>(UPLOAD_PHOTO);
+  // Uploaded Image File
+  const [imageURLFile, setImageURLFile] = useState(null); // Path of uploaded image on client, to be used in image previews
```
cristian-aldea: You missed the easiest one xD

```suggestion
  const [imageURLFile, setImageURLFile] = useState<string>(null); // Path of uploaded image on client, to be used in image previews
```

apps/mull-ui-e2e/src/integration/create-event.spec.ts
cristian-aldea, line 1: Actually I just remembered, didn't we agree that we would make e2e tests, one for each user story?

Ideally, we should find a way to split this file into two: `US-1.1.spec.tsx` and `US-1.2.spec.tsx`.

However, I know that 1.2 is more of a supporting US for 1.1, and that their flow is quite intertwined, so I suggest that **at least** you indicate in this file that it covers both user stories. Both by renaming it to something like `US-1.1.US-1.2.spec.tsx` and changing the `describe` statement on line 10.

apps/mull-ui/src/app/pages/create-event/date-calendar/date-calendar.scss
cristian-aldea, line 4: I see you added the styling to each class in create.event.scss. However now this line you added is dead code since its applied twice in two places. Please remove it.

cristian-aldea (changes requested):
Looking good! Had a few final comments that I wanted addressed. One is about the e2e test, which is why I set my review as a blocker. The others are non-blocking, but would still like to see addressed.

apps/mull-ui/src/app/components/custom-file-upload/custom-file-upload.scss
r-saba: I'd prefer to have another task where I clean up the UI and CSS since I'm not sure what else there is and the UI can use some refining to better match our mockups.  Let me know what you think and I'll add it to the backlog if you agree.

apps/mull-ui-e2e/src/integration/create-event.spec.ts
r-saba: I remember it was brought, don't remember the resolution we agreed on. The e2e for `US-1.2` is tightly coupled in at least one of the e2e tests for `US-1.1`. I will change the file name and `describe` statement for now we can discuss with team what we will do going forward.

apps/mull-ui/src/app/components/custom-file-upload/custom-file-upload.scss
cristian-aldea: yup, its better as a seperate bug or fix pr

apps/mull-ui-e2e/src/integration/create-event.spec.ts
r-saba: True

r-saba:
After discussion with the team US-1.2 is now a TASK for [US-1.1](https://github.com/RGPosadas/Mull/issues/27). The documentation will be updated once this PR is merged. We are keeping this PR instead of making another one since there is important discussion here we don't want to lose.

apps/mull-ui/src/app/pages/create-event/create-event.tsx
```diff
@@ -44,16 +48,19 @@ const CreateEventPage = ({ history }: CreateEventProps) => {
...
   const toastId = useRef(null);
   // GraphQL mutation hook to create events
-  const [createEvent] = useMutation(CREATE_EVENT);
-  // Image of Event
-  const [imageFile, setImageFile] = useState(null);
-
+  const [createEvent] = useMutation<IEvent>(CREATE_EVENT);
+  const [uploadFile] = useMutation<{ uploadFile: IMedia }>(UPLOAD_PHOTO);
+  // Uploaded Image File
+  const [imageURLFile, setImageURLFile] = useState<String>(null); // Path of uploaded image on client, to be used in image previews
```
cristian-aldea:
```suggestion
  const [imageURLFile, setImageURLFile] = useState<string>(null); // Path of uploaded image on client, to be used in image previews
```
https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#number-string-boolean-symbol-and-object

apps/mull-ui/src/app/pages/create-event/date-calendar/date-calendar.scss
TheGreatMarkus: This line is still here

cristian-aldea (dismissed):
I won't block this since I know you'll be busy. Just make sure to adress the comment I didn't resolve:
https://github.com/RGPosadas/Mull/pull/77/files#r528244459

- [x] User acceptance flow is respected
- [x] Documentation present where needed
- [x] There must exist an automated Cypress acceptance test
- [x] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=2000%3A222910)

apps/mull-ui/src/app/pages/create-event/date-calendar/date-calendar.scss
r-saba: I removed the line in `create.event.scss` and kept it in `date-calendar.scss` it made more sense to style the class in `date-calendar.scss`.

cristian-aldea (approved):

JimLau49 (approved):

bit172 (approved):

Issue #70, opened 2020-11-09
TASK-17: Event Page


cristian-aldea:
**Task Description**

As a developer, I want a event page so that user can look at the details of an event.

**Extra Description**

This page will be used in multiple places in the application, including when the user joins an event (#66), and when the user reviews an event they are about to create (#48). This task aims to support such user stories.

**Subtasks**

~~17.1 Back button~~ Now a separate task #78
- [x] 17.2 Event page header
- [x] 17.3 Event info
- [ ] 17.4 Mull button
- [ ] 17.5 Integrate Components into this task.

**Acceptance Criteria**

- [ ] Documentation present where necessary
- [ ] Unit and UI tests are implemented for every component and function
- [ ] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=2007%3A404529)

r-saba:
This ticket will have 4 subtasks.
1. Back button
2. Event page header
3. Event Info
4. Mull Button

cristian-aldea:
Created branch `TASK-17/event-page` on which to integrate the changes from the subtasks

cristian-aldea:
I'll be starting work on 17.1 on `TASK-17.1/back-button`

cristian-aldea:
Something I'll bring up in the next scrum, I think maybe the back button should be a bit different.
![image](https://user-images.githubusercontent.com/23544999/98858168-28690980-242e-11eb-9876-003b7c295431.png)

If we compare this to back buttons in other mobile apps (Youtube and Instagram):
![image](https://user-images.githubusercontent.com/23544999/98858416-91508180-242e-11eb-9200-7d708e10cb8c.png)
![image](https://user-images.githubusercontent.com/23544999/98858443-99a8bc80-242e-11eb-8202-366e1fc4d526.png)

We can see that in other apps, Instead of the back button being part of the page, it's part of the header. And instead of the text next to it being the previous page, the text is usually the current page title.

I think for now we can keep the back button on hold until we discuss this

cristian-aldea:
Starting work on 17.2 on `TASK-17.2/event-page-header`

cristian-aldea:
I'll be taking care of both 17.4 and 17.5 and get a PR up soon

PR #79, opened 2020-11-14
Create event page header component #70

cristian-aldea:
Added some necessary changes to backend entities to be able to use event interface in front-end


cristian-aldea:
No need for a pr for subtasks, since ritchelle and I are working closely together on #70

Issue #78, opened 2020-11-13
TASK-22: Back Button


r-saba:
**Task Description**

As a developer I want a reusable back button component that I can use on any page in the app.

**Acceptance Criteria**

- [ ] Documentation present where necessary
- [ ] Unit and UI tests are implemented for every component and function
- [ ] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=533%3A6528)

r-saba:
Marked as waiting for refinement - This ticket needs to be estimated, and we also need to clarify exactly what is involved in this ticket. Is it just front-end? Does it handle routing to the previous page?

PR #80, opened 2020-11-15
TASK-17: Event Page

cristian-aldea:
This PR addressed #70 

This PR aims to implement an event page for use in other user stories.

The images used in the component are placeholders until saving to and querying from the media server is done.

You can access the page by going to http://localhost/test-event-page

Notes:
* Added interfaces for events and related entities, and made the existing entities inherit those interfaces. That way the front-end can use these interfaces.
* Adjusted styling of html, body, and container divs to take up the entire width and height of the page. Without this, the scrolling in the event page info would be much less feasible. 
* Fixed useless test in apps/mull-ui/src/app/components/header/header.spec.tsx
* Moved button from create event page into its own component, under a components/ folder
* created EventRestrictionMap for easy access to enum string values
* Fixed inter font not being loaded due to incorrect url
* although this issue is blocked by #78, I created the PR to get feedback


r-saba (changes requested):
Looking great! Here's my UI comments

- The green horizontal bar between event info fields is too dark compared to figma
- Missing the last green horizontal bar, there's one under each info row
- The icons on the left are too large
- The event description should have "view more" and once we click that we see the entire text
- Missing the last row, "invite friends"

cristian-aldea:
Just a note @r-saba, please try to not have too many comments in the main review text, I did that mistake before and it makes it hard to reply (since I can only quote reply)

Here is my response: 
> * The green horizontal bar between event info fields is too dark compared to figma

Sure thing will fix

> * Missing the last green horizontal bar, there's one under each info row

If you look at the mockup you'll see that the divider lines are a bit inconsistent
image
I tried seeing how it would look if I added lines for the top of the first row and the bottom of the last row, but I didn't like how it looked. We can discuss the specific styling and what we should go for in a meeting

> * The icons on the left are too large

True, ill fix that

> * The event description should have "view more" and once we click that we see the entire text

Ritchelle and I discussed and we felt that that might not be necessary, I can implement it real quick if we need its needed.

> * Missing the last row, "invite friends"

True, didn't include it at the time since a few details we're still shaky, will add it.

r-saba:
@TheGreatMarkus Sure will make single comments next time.
Regarding the missing divider at the bottom, do whatever you feel looks nice.
Regarding the "view more", I do think it's necessary. Unless we limit the number of characters in the event description, it could end up taking too much space on this page. If you could implement real quick that'd be sweet thanks!

cristian-aldea:
@r-saba sounds good, I'll implement the read more thing, and as for the dividers I can ask that real quick during our scrum

cristian-aldea:
@r-saba implemented the expandable text as a separate component, I'm quite happy with how it turned, since its very flexible and easy to use imo. Lemme know what you think, I've addressed all your comments (aside from the div lines, which we'll talk during the meeting today), so I've asked for your review again.

RGPosadas:
Looked over the UI and I don't have any comments other than good job, it looks and works great! I am just waiting for #78 to be merged so that I can give you a full review.

apps/mull-ui/src/app/components/event-page-info/event-page-info.scss
```diff
@@ -0,0 +1,30 @@
...
+  display: flex;
+  flex-direction: row;
+  padding: 0.5rem 1rem;
+  box-sizing: border-box;
+  align-items: flex-start;
+}
+
+.event-page-info-container {
+  div {
+    border-bottom: 1px solid rgba($color: $primary-color, $alpha: 0.5);
```
r-saba: Matches better the mockup and the mockups use 0.5 for width.
```suggestion
    border-bottom: 0.5px solid rgba($color: $primary-color, $alpha: 0.5);
```

apps/mull-ui/src/app/components/event-page-header/event-page-header.scss
```diff
@@ -0,0 +1,49 @@
+.event-page-header {
+  width: 100%;
+  padding: 0 0.5rem;
+  box-sizing: border-box;
+  box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.1);
+}
+
+.title {
+  font-size: 1.5em;
+  font-weight: bold;
```
r-saba: Title is not bolded in the mockup. 

ps - Since this is on a h1, and we haven't cleared the styles the `font-weight: bold` is unnecessary since it's bold by default styling.

apps/mull-ui/src/app/components/event-page-header/event-page-header.scss
```diff
@@ -0,0 +1,49 @@
...
+  width: 100%;
```
r-saba: We're not using flex, by default it will take up the entire space.
```suggestion
```

apps/mull-ui/src/app/components/event-page-header/event-page-header.scss
```diff
@@ -0,0 +1,49 @@
...
+  text-align: center;
+  overflow-y: auto;
```
r-saba: Is there a reason for this?

apps/mull-ui/src/app/components/event-page-header/event-page-header.scss
```diff
@@ -0,0 +1,49 @@
...
+  max-height: 3em;
```
r-saba: Is there a reason for this?

apps/mull-ui/src/app/components/event-page-header/event-page-header.scss
```diff
@@ -0,0 +1,49 @@
...
+  box-sizing: border-box;
+  margin: 0.5rem 0;
+}
+
+.event-image {
+  max-width: 100%;
```
r-saba: Typically we use max-width when we want something to stay below 100%. `max-width: 50%`.
This isn't necessary here.
```suggestion
```

apps/mull-ui/src/app/components/event-page-header/event-page-header.scss
```diff
@@ -0,0 +1,49 @@
...
+  border-radius: 8px;
+  box-sizing: border-box;
```
r-saba: I don't typically use `border-box` why is this everywhere? When I remove it in the inspector nothing changes

apps/mull-ui/src/app/components/event-page-info/event-page-info.scss
```diff
@@ -0,0 +1,30 @@
+@import 'variables';
+
+.info-row {
+  width: 100%;
+  display: flex;
+  flex-direction: row;
+  padding: 0.5rem 1rem;
+  box-sizing: border-box;
+  align-items: flex-start;
+}
```
r-saba: - `flex-direction` is row by default not need to explicitly define it.
- row will be width 100% by default no need to explicitly define it.
- Do we need `align-items: flex-start`? I don't usually have to use this.

libs/ui-lib/src/components/mull-button/mull-button.scss
```diff
@@ -0,0 +1,12 @@
+@import '../../lib/styles/variables.scss';
+
+.mull-button {
+  color: white;
+  font-weight: bold;
+  padding: 1rem 3rem;
+  background: $primary-color;
+  border-radius: 9px;
+
+  border: none;
```
r-saba: Button doesn't have a border by default anyway in our app.
```suggestion
.mull-button {
  color: white;
  font-weight: bold;
  padding: 1rem 3rem;
  background: $primary-color;
  border-radius: 9px;
```

libs/ui-lib/src/components/expandable-text/expandable-text.tsx
```diff
@@ -0,0 +1,41 @@
...
+          {expand ? ' ' : '... '}
+          <button
+            data-testid="expandable-text-button"
+            className="expandable-span-button"
+            onClick={() => setExpand(!expand)}
+          >
+            {expand ? 'less' : 'more'}
+          </button>
+        </>
+      )}
```
r-saba: I would move this into a function and call the function here. Typically we don't want "complex" logic inside the return statement.

libs/ui-lib/src/lib/styles/styles.scss
```diff
@@ -11,6 +11,13 @@
   border: 0;
 }
 
+html,
+body,
+#root {
+  width: 100%;
+  height: 100%;
+}
+
```
r-saba: I don't like this, have you considered how it will affect other styling? 
I see you're doing this to keep the button at the bottom I know there was a discussion on if we're keeping it sticky or part of the scrolling section. If it needs to stick to the bottom, consider position abolsute.

r-saba (changes requested):
Since the PR is strictly front-end I'm paying a lot more attention to the CSS. Just be careful with what you add. CSS that has not been properly thought through can cause a lot of head ache for the next developer when their elements don't style as expected. Also unnecessary/dead CSS is confusing. When in doubt steer towards specificity.

We typically don't have to specify height/width 100% unless we're using something like flex or some other styling is affecting our component.

- Let's replace `em` with `rem` to be consistent. I noticed you were using both in this pr.
- Double check all your CSS

apps/mull-ui/src/app/components/event-page-header/event-page-header.scss
cristian-aldea: You're right, ill make it a div and remove the font weight

cristian-aldea:
thanks @r-saba for giving this so much attention, will look into your comments. Just had some quick replies to your comments:

> * Let's replace `em` with `rem` to be consistent. I noticed you were using both in this pr.

We can both use `em` and `rem` when suitable, no needed to force ourselves to use one over the other. I do admit some of my uses of `em` were accidental/bad, but just felt like pointing that out.

> * Double check all your CSS

will do o7

apps/mull-ui/src/app/components/event-page-header/event-page-header.scss
cristian-aldea: I set a max height for the text of 3 lines (3em) to account for long titles. If the title ends up longer than that I make it scrollable.

apps/mull-ui/src/app/components/event-page-header/event-page-header.scss
cristian-aldea: See above

apps/mull-ui/src/app/components/event-page-header/event-page-header.scss
cristian-aldea: Removing this

apps/mull-ui/src/app/components/event-page-header/event-page-header.scss
cristian-aldea: Removing this

cristian-aldea: I like to use box-sizing: border-box when I center since otherwise things get misaligned sometimes due to padding. Not necessary in this case.

apps/mull-ui/src/app/components/event-page-info/event-page-info.scss
cristian-aldea: Thanks for pointing that out, I'm not familiar with all the default, will be removing these lines

apps/mull-ui/src/app/components/event-page-info/event-page-info.scss
cristian-aldea: Yup, will also simplify this, since we don't need to exclude the last children from having a border

libs/ui-lib/src/components/expandable-text/expandable-text.tsx
cristian-aldea: You would move the tsx here to a function? I don't feel like that's necessary. lets discuss about this together

libs/ui-lib/src/components/mull-button/mull-button.scss
cristian-aldea: I think this was left over due to jimmy and jon putting a border for some reason on the original button I moved out of create event.

I probably set it to none just in case, even though its not needed, much like the rest of the dead css in the rest of this pr haha

libs/ui-lib/src/lib/styles/styles.scss
cristian-aldea: This isn't to have the button stuck at the bottom. This is to make the event info component scroll correctly doable. For the event info to scroll but not the event page itself, the event page needs to take up the whole page, which wasn't possible since `html`, `body` and `#root` had `fit-content` as their height.

I really doubt this will negatively affect any styling imo, since all it does is make sure that containers take up the entire page, which I thought it was doing already. Also I tested the existing page and no stylng broke as far as I know. 

This warrants a longer conversation, thought, and obviously I might be wrong on what I said earlier. Open to alternatives since you have more experience than me on this.

apps/mull-ui/src/app/components/event-page-header/event-page-header.scss
cristian-aldea: As we discussed, we will assume title aren't absurdly long

apps/mull-ui/src/app/components/event-page-header/event-page-header.scss
cristian-aldea: caused by the width:100% in app.scss. Removing all/most of my uses of border-box since its no longer necessary

libs/ui-lib/src/lib/styles/styles.scss
cristian-aldea: found an alternative solution. The height of the page is 100vh - height of navbar - height of header

apps/mull-ui/src/app/components/event-page-info/event-page-info.scss
```diff
@@ -0,0 +1,28 @@
+@import 'variables';
+
+.event-page-info-container {
+  display: flex;
+  flex-direction: column;
+}
+
+.info-row {
+  display: flex;
+  padding: 0.5rem 1rem;
```
r-saba: I find the rows could use more breathing room. The spacing in the mockups is not consistent so it's hard to say exactly what we should use. I suggest 1.25rem.
```suggestion
  padding: 1.25rem 1rem;
```

apps/mull-ui/src/app/components/event-page-info/event-page-info.scss
```diff
@@ -0,0 +1,28 @@
...
+  border-bottom: 0.5px solid rgba($color: $primary-color, $alpha: 0.5);
+}
+
+.event-page-button {
+  align-self: center;
+  margin: 0.5rem 0;
```
r-saba: Button could use more margin-top to match the mockup and look better.

r-saba (changes requested):
Thanks for addressing all my comments. Really good work! I left just a very minor feedback on the UI below.

libs/ui-lib/src/lib/styles/styles.scss
r-saba: Nice! That's what I was getting at with `vh` during our discussion, very useful!

r-saba:
Regarding `em` vs `rem`. I haven't seen the usage of `em` at all actually in web apps. rem is a standard and I suggest we stick with it unless we really have a specific use case to use `em`. Using both will result in inconsistencies throughout our app.

cristian-aldea:
> Regarding `em` vs `rem`. I haven't seen the usage of `em` at all actually in web apps. rem is a standard and I suggest we stick with it unless we really have a specific use case to use `em`. Using both will result in inconsistencies throughout our app.

I disagree, from reading articles like https://zellwk.com/blog/rem-vs-em/, each have their usecase. `em` is also a standard, just like `rem`. It's true that so far we haven't had much of a usecase for `em` in our code, but I just don't want the use of `rem` to be dogmatic.

For example, if we end up needing to define bottom padding for different font sizes (p, h1, h2, h3, etc.) that would be a perfect use case for em, since we'd want the padding to change depending on the size of the current element's text.

cristian-aldea:
Sorry for all the pushes @r-saba, i tried by best to clean up the git history

apps/mull-ui/src/app/components/event-page-info/event-page-info.scss
cristian-aldea: Nice catch, I do agree that some more padding is needed, however I feel that 1.25 is excessive from my testing. I bumped up the padding to 1rem:

implementation:
image

mockup:
image

apps/mull-ui/src/app/components/event-page-info/event-page-info.scss
cristian-aldea: Yup agree, top and bottom should have more margin

r-saba:
As I mentioned I haven't seen `em` used. Even the website you linked is using rem. Also I've found [plenty ](https://webdesign.tutsplus.com/tutorials/comprehensive-guide-when-to-use-em-vs-rem--cms-23984) of articles suggesting exactly what I said, the usage of `rem` since it's always consistent and using `em` only if it is necessary. 

I haven't worked with design systems that spaces content based on font size as you suggest. Typically we have a defined spacing for layout (40px b/w headings and content, 10px between content and buttons etc..). We're saying the same thing that `em` has some use cases in sepcific situations (though I've looked the css of many website over the years and haven't seen it used). However we should use `rem` where we can since it's behavior is consistent and easier to predict, and use `em` only if we find a good use case for it. Using both liberally will make our UI look unprofessional.

apps/mull-ui/src/app/components/mull-back-button/mull-back-button.tsx
```diff
@@ -1,33 +1,35 @@
...
+  const defaultOnClick = () => {
+    history.goBack();
+  };
+
   return (
     <button
       data-testid="mull-back-button"
       className={`mull-back-button ${className}`}
-      onClick={onClick}
+      onClick={onClick ? onClick : defaultOnClick}
```
r-saba:
```suggestion
      onClick={onClick || defaultOnClick}
```

apps/mull-ui/src/app/components/event-page-header/event-page-header.scss
```diff
@@ -0,0 +1,38 @@
...
+.event-image {
+  border-radius: 8px;
+  display: block;
+  max-height: 10rem;
+  margin: 0 auto;
+}
+
+.event-datetime {
+  display: flex;
+  flex-direction: row;
```
r-saba:
```suggestion

```

apps/mull-ui/src/app/components/event-page-header/event-page-header.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+import './event-page-header.scss';
+
+import { IEvent } from '@mull/types';
+
+import MullBackButton from '../mull-back-button/mull-back-button';
+
+export interface EventPageHeaderProps {
+  event: IEvent;
+  lastPage: string;
```
r-saba: can you rename to `prevPage`? `lastPage` sounds like it's related to pagination

apps/mull-ui/src/app/components/mull-button/mull-button.scss
```diff
@@ -0,0 +1,11 @@
+@import '../../lib/styles/variables.scss';
+
+.mull-button {
+  color: white;
+  font-weight: bold;
+  padding: 1rem 3rem;
+  background: $primary-color;
+  border-radius: 9px;
+
```
r-saba: Unnecessary spacing
```suggestion
```

apps/mull-ui/src/app/pages/event-page/event-page.tsx
```diff
@@ -0,0 +1,23 @@
+import { IEvent } from '@mull/types';
+import React from 'react';
+import { EventPageHeader, EventPageInfo } from '../../components';
+
+import './event-page.scss';
+
+export interface EventPageProps {
+  event: IEvent;
+  lastPage: string;
```
r-saba: If you change it to `prevPage` above, change it here too

r-saba (dismissed):
LGTM! None of my comments below are blockers. If you do address these comments let me know. If you tell me you didn't touch any other part of the code I will approve it quickly after testing functionality.

apps/mull-ui/src/app/components/mull-back-button/mull-back-button.tsx
cristian-aldea: I had a feeling there was a better way of doing this. Thanks!

cristian-aldea:
@r-saba Yeah we're basically saying the same thing, and I admit my uses of `em` in earlier sprints and in this pr were mostly due to ignorance or error. Just wanted to make sure people reading this understand why we use one over the other.

RGPosadas (dismissed):
**Subtasks**
- [x] 17.2 Event page header
- [x] 17.3 Event info
- [x] 17.4 Mull button
- [x] 17.5 Integrate Components into this task.

**Acceptance Criteria**
- [x] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=2007%3A404529)

**Comments**
Great job! The added padding between the div and the row content (compared to the first time I reviewed this PR) is a great subtle addition.
LGTM!

apps/mull-ui/src/app/components/event-page-header/event-page-header.tsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faCaretRight } from '@fortawesome/free-solid-svg-icons';
+
+import { faClock } from '@fortawesome/free-regular-svg-icons';
+
```
JimLau49: I think we can have svg imports without spaces in between them

JimLau49 (dismissed):
LGTM! , really nice work you have here, i left a small comment as single comment instead of a full review that could be addressed. 

Additionally, ive noticed the caret-right, we're using from FA doesn't exactly match what we have in the mockup.

TASK-17
image

Mockup:
image

Im not sure if there is a similar one offered in FA, if not a small edit to the mockup should be done at some point.

Great work 👍

cristian-aldea:
@JimLau49 
Yeah unfortuantely searching for icons I couldn't find one that matched the mockup. It doesn't really matter anyways

thanks, took a lot longer than I thought!

apps/mull-ui/src/app/components/event-page-header/event-page-header.tsx
cristian-aldea: yup, removed the space

cristian-aldea:
Alright I addressed all the comments I saw, lemme know if you see anything else.

JimLau49 (approved):
Review remains as an approve

RGPosadas (approved):

r-saba (approved):
LGTM! Great work

PR #82, opened 2020-11-16
TASK-22: Back button

cristian-aldea:
This PR addressed  #78

Notes:
* 9f4abc0 adds a temporary button in create event page so people can test as part of this pr, will be removed before merging.
* e2e tests are failing, but that is normal since I modified the create event page to test the back button. e2e tests will work again when test back button is removed from create event page

To see and test the button:
* Go to a page such as the Create Event page: `apps/mull-ui/src/app/pages/create-event/create-event.tsx`
* import the button `import { MullBackButton } from '../../components';`
* Add it somewhere in the page: `<MullBackButton history={history}>Back Button!!</MullBackButton>`

### DIFF
--- a/libs/ui-lib/src/lib/custom-time-picker/custom-time-picker.spec.tsx
+++ b/libs/ui-lib/src/lib/custom-time-picker/custom-time-picker.spec.tsx
@@ -20,13 +20,6 @@ describe('CustomTimePicker', () => {
     expect(baseElement).toBeTruthy();
   });
 
-  it('should render successfully', () => {
-    const props = mockCustomTimePickerProps();
-    props.hasErrors = false;
-    const { baseElement } = render(<CustomTimePicker {...props} />);
-    expect(baseElement).toBeTruthy();
-  });
-
   it('should match snapshot', () => {
     const props = mockCustomTimePickerProps();
     const tree = renderer.create(<CustomTimePicker {...props} />).toJSON();
--- a/apps/mull-ui/src/app/app.tsx
+++ b/apps/mull-ui/src/app/app.tsx
@@ -3,8 +3,7 @@ import { ToastContainer, toast } from 'react-toastify';
 import { Redirect, Route, Switch } from 'react-router-dom';
 
 import CreateEventPage from './pages/create-event/create-event';
-import NavigationBar from './components/navigation-bar/navigation-bar';
-import Header from './components/header/header';
+import { NavigationBar, Header } from './components';
 import { ROUTES } from '../constants';
 
 import 'react-toastify/dist/ReactToastify.min.css';
--- a/apps/mull-ui/src/app/components/header/header.spec.tsx
+++ b/apps/mull-ui/src/app/components/header/header.spec.tsx
@@ -19,6 +19,8 @@ describe('Header', () => {
         <Header />
       </Router>
     );
+
+    expect(baseElement).toBeTruthy();
   });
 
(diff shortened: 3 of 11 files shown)


libs/ui-lib/src/lib/custom-time-picker/custom-time-picker.spec.tsx
```diff
@@ -20,13 +20,6 @@ describe('CustomTimePicker', () => {
     expect(baseElement).toBeTruthy();
   });
 
-  it('should render successfully', () => {
```
RGPosadas, line 23: I see you've taken out a duplicate unit test, so nice catch!
However, please state that you had refactored in the commit as well (and also add in the `refactor` label on your PR) 😄

RGPosadas (commented):
Sizing and current styling of the button looks good! Just had a few questions/comments:

* I have a comment about placement. Since this will be a component that will have its position locked on mobile, should styling be added so it just stays in the upper left corner? 
* I also remembered that for desktop, we will be using the browser's back button so this will not be used in desktop view. Should the logic of the button disappearing in desktop view be added here too (frontend nub pls halp)?

libs/ui-lib/src/lib/custom-time-picker/custom-time-picker.spec.tsx
cristian-aldea: I'll be cleaning up my commits as I rebase

cristian-aldea:
> * I have a comment about placement. Since this will be a component that will have its position locked on mobile, should styling be added so it just stays in the upper left corner?

We could, but I don't think that's necessary, the page (or parent component) which uses this back button should control where it is positionned. The button itself should control its children (the caret and the text), how they look, and where they are. in relation to each other.

> * I also remembered that for desktop, we will be using the browser's back button so this will not be used in desktop view. Should the logic of the button disappearing in desktop view be added here too (frontend nub pls halp)?

Again, the page which uses this button should implement that imo.

RGPosadas:
> > * I have a comment about placement. Since this will be a component that will have its position locked on mobile, should styling be added so it just stays in the upper left corner?
> 
> We could, but I don't think that's necessary, the page (or parent component) which uses this back button should control where it is positionned. The button itself should control its children (the caret and the text), how they look, and where they are. in relation to each other.
> 
> > * I also remembered that for desktop, we will be using the browser's back button so this will not be used in desktop view. Should the logic of the button disappearing in desktop view be added here too (frontend nub pls halp)?
> 
> Again, the page which uses this button should implement that imo.

I see, thank you for the explanation!

RGPosadas (dismissed):
**Acceptance Criteria**

- [x] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=533%3A6528)

cristian-aldea:
@RGPosadas sorry I removed my temporary commit (I should let the reviewer make those change so that I won't need to reask for an approval) and added a new refactor

RGPosadas (approved):
My previous approval still stands. LGTM!

cindyslittleplanet (approved):
**Acceptance Criteria**
- [x] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=533%3A6528)

Just want a double check before you merge, the current code coverage is under 70% because of the failing E2E tests correct? If yes, you can go ahead and merge.

PR #83, opened 2020-11-16
FIX: Add font awesome and remove some svgs

cristian-aldea:
This is currently blocking #82 from being merged. For reasons why see #82


JimLau49 (approved):
LGTM 

This will make our svg usage a bit easier 🚀

r-saba (approved):
LGTM! By using font awesome we have more flexibility with our icons, and we don't have to worry about optimizing them for screen sizes. Thanks @TheGreatMarkus

Issue #72, opened 2020-11-11
TASK-18: Multiple Registration Options Front-end 


r-saba:
**Task Description**

As a developer I want a UI to allow users to create an account using multiple registration options.

**Subtasks**
- [x] 18.1 Login page
- [x] 18.2 Registration page

**Acceptance Criteria**

- [ ] Documentation present for every new component and function where necessary
- [ ] Unit and UI tests are implemented for every component and function
- [ ] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=2079%3A70811)

Issue #30, opened 2020-09-27
US-7.1: Multiple Registration Options


RGPosadas:
**User Story Description**

As a user, I want to have multiple registration options so that I have the choice of registering with an existing service (Google, etc) or through email.

**User Acceptance Flow**

1. User opens the app
1. User sees multiple ways of registering
    a. Through an existing service (Google, Twitter, Facebook, etc)
    b. Using user's email
1. User chooses an option to make a Müll account

**Acceptance Criteria**

- [ ] User acceptance flow is respected
- [ ] Registration page is implemented
- [ ] Documentation present for every new component and function
- [ ] Unit and UI tests are implemented for every component and function
- [ ] There must exist an automated Cypress acceptance test
- [ ] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=2079%3A70811)

PR #84, opened 2020-11-16
TASK-18: multiple registrations options front end

JimLau49:
This PR implements the front-end of Multiple Registrations Options #72 login page and registration page as well as their jest unit tests.

You can access the pages by going on `http://localhost:4200/login` and `http://localhost:4200/register` 

Notes: 
Added a field in custom-text-input for password fields 
Header and navigation bar should not be visible while on these pages.


apps/mull-ui/src/app/app.tsx
```diff
@@ -13,6 +15,14 @@ import './app.scss';
 /**
  * Main component of the application
  */
+const showHeader = () => {
+  const location = useLocation();
+  if (['/login', '/register'].includes(location.pathname)) {
```
cristian-aldea: Please change this line to
```suggestion
  if ([ROUTES.LOGIN, ROUTES.REGISTER].includes(location.pathname)) {
```

and change the routes enum to an object:

```ts
export const ROUTES = {
  HOME: '/home',
  MAP: '/map',
  CREATE_EVENT: '/create-event',
  TOOLS: '/tools',
  MESSAGES: '/messages',
  PROFILE: '/profile',
  LOGIN: '/login',
  REGISTER: '/register',
};
```

I tested it and thankfully nothing breaks when you do that

apps/mull-ui/src/app/app.tsx
```diff
@@ -13,6 +15,14 @@ import './app.scss';
 /**
  * Main component of the application
  */
+const showHeader = () => {
```
cristian-aldea: Maybe rename this function to `showNavigation`?

apps/mull-ui/src/app/app.tsx
```diff
@@ -13,6 +15,14 @@ import './app.scss';
...
+  const location = useLocation();
```
cristian-aldea, line 23: `useLocation` is a hook, so it's tied to the state of a react component. You can include it in the App component as a field, and pass it to the function when calling it, that way you don't redefine the hook every time the function is called.

apps/mull-ui/src/app/components/navigation-bar/navigation-bar.spec.tsx
```diff
@@ -41,11 +41,11 @@ describe('NavigationBar', () => {
     );
 
     for (const key in ROUTES) {
-      history.push(ROUTES[key]);
-
-      const element = dom.getByTestId(testIds[ROUTES[key]]);
-
-      expect(element.classList.contains('active')).toBeTruthy();
+      if (!['LOGIN', 'REGISTER'].includes(key)) {
```
cristian-aldea: To not have to change this code every time we introduce a new route , you should instead define an array with ONLY the routes we wanna test: (HOME, MAP, CREATE_EVENT, TOOLS, PROFILE, MESSAGES), instead of specifying the list of routes we DONT want to check (LOGIN, REGISTER).

Since I wrote this test I can update it if you're confused as to what to do

apps/mull-ui/src/app/pages/login/login.scss
```diff
@@ -0,0 +1,136 @@
+@import 'variables';
+
+.login-container {
+  display: grid;
+  grid-template-columns: 1fr;
+  grid-column-gap: 1.5em;
+  grid-row-gap: 1.5em;
+  margin: 0 auto;
+  width: 95%;
```
cristian-aldea: Instead of setting the width to 95%, you can set horizontal padding. Putting padding would allow you to remove the `margin: 0 auto;` line too. Regardless of what you decide to do, make the width a bit smaller (or the padding a bit bigger) to match the mockup.

apps/mull-ui/src/app/pages/login/login.scss
```diff
@@ -0,0 +1,136 @@
+@import 'variables';
+
+.login-container {
```
cristian-aldea, line 3: There is a big space at the top of the page on the mockup which is missing in the implementation.
Please add a top padding for this container

libs/ui-lib/src/lib/custom-text-input/custom-text-input.tsx
```diff
@@ -31,21 +32,36 @@ export const CustomTextInput = ({
...
+        <input
+          className={`custom-text-input input-border ${hasErrors ? 'error' : ''}`}
+          id={fieldName}
+          name={fieldName}
+          data-testid="custom-text-input"
+          type="text"
+          value={value}
+          onChange={onChange}
+        />
+      )}
```
cristian-aldea: This is a really bad way of doing it. What I would recommend is doing what you did for the other attributes such as id, name, value and onChange. Just create a "type" props for the CustomTextInput component, and pass it when directly to the `<input>` element you have here. Its going to make your code much simpler.

When you're making a custom `<input>` component like in this case, you essentially want to make a wrapper around a `<input>` element, and let the client code give the values it wants to set to the wrapped `<input>` element.

You can set a type of `"text" | "password" | "something else ..."` to the prop if you want

cristian-aldea (changes requested):
The login and register pages look great! Had a few comments, however.

Also, when you rebase, make sure to use the font awesome icons and remove your svgs. We can be flexible and change the mockup if we don't find the exact styling of icons we want.

apps/mull-ui/src/app/app.tsx
JimLau49: Great catch cristian, will do 👍

apps/mull-ui/src/app/pages/login/login.scss
JimLau49: I wasn't here for this change, tagging @leojrsilao

libs/ui-lib/src/lib/custom-text-input/custom-text-input.tsx
JimLau49: Sounds good, we were debating between either doing the prop or doing it using spread operator as above since we didn't think we needed anything more than text and password, but we can make the change you've suggested

apps/mull-ui/src/app/components/navigation-bar/navigation-bar.spec.tsx
JimLau49: We'll try to fix it, ill reach out if we get stuck on updating this test

apps/mull-ui/src/app/pages/login/login.scss
leojrsilao: Will change it to have horizontal padding

apps/mull-ui/src/app/pages/login/login.scss
leojrsilao: Will add the top padding for the container.

libs/ui-lib/src/lib/custom-text-input/custom-text-input.tsx
cristian-aldea: Even if we won't need a type other than text or password, not having an if else if going to make the code easier to maintain

cristian-aldea:
Just to clarify, make sure to install the brand style for fontawesome: https://www.npmjs.com/package/@fortawesome/free-brands-svg-icons to have icons for facebook, google, etc.

apps/mull-ui/src/app/app.tsx
```diff
@@ -21,6 +31,8 @@ export const App = () => {
           {<Redirect to={ROUTES.HOME} />}
         </Route>
         <Route exact path={ROUTES.CREATE_EVENT} component={CreateEventPage} />
+        <Route exact path={ROUTES.LOGIN} component={LoginPage} />
+        <Route exact path={ROUTES.REGISTER} component={RegisterPage} />
```
bit172, line 45: Since the login/register page should appear based on whether the user has logged in or not. What I would do, is to have a user object in either a React Context or in memory (for now until we get #30 started) set to null and do some authentication checks for the routes. It would require a change in our current app component so I'm not sure if this is within the scope of this ticket but it needs to be done in #30. Here's an [example](https://reactrouter.com/web/example/auth-workflow) from the React Router docs.

apps/mull-ui/src/app/app.tsx
```diff
@@ -13,6 +15,14 @@ import './app.scss';
 /**
  * Main component of the application
  */
+const showHeader = () => {
+  const location = useLocation();
+  if (['/login', '/register'].includes(location.pathname)) {
+    return false;
+  }
+  return true;
+};
```
bit172: I do not think this is a good way of handling the UI. The route has nothing to do with whether we can see the Header or not. My comment below goes into greater detail. In short, use authentication, not URL paths.

apps/mull-ui/src/app/pages/login/login.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+export const Login = ({ history }) => {
+  const formik = useFormik({
+    initialValues: {
+      email: '',
+      password: '',
+    },
+
+    validationSchema: Yup.object({
+      email: Yup.string().required('Email is required.'),
```
bit172: You could also add a valid email check here. Yup does come with that [here](https://github.com/jquense/yup#stringemailmessage-string--function-schema).

apps/mull-ui/src/app/pages/register/register.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const formik = useFormik({
+    initialValues: {
+      name: '',
+      email: '',
+      password: '',
+    },
+
+    validationSchema: Yup.object({
+      name: Yup.string().required('Name is required.'),
+      email: Yup.string().required('Email is required.'),
```
bit172: Email check here too.

libs/ui-lib/src/lib/custom-text-input/custom-text-input.tsx
bit172: Yeah, feel free to change up the custom components as requirements evolve. I would suggest doing this since it would probably be text/password only:
```suggestion
        <input
          className={`custom-text-input input-border ${hasErrors ? 'error' : ''}`}
          id={fieldName}
          name={fieldName}
          data-testid="custom-text-input"
          type={password ? "password": "text"}
          value={value}
          onChange={onChange}
        />

```
`password` would be a prop and you would use it like this:

Password:
```jsx
        <CustomTextInput
          title="Password"
          fieldName="password"
          value={value}
          onChange={handleChange}
          hasErrors={checkFunction}
          errorMessage={errorMessage}
          password
        />
```
Text:
```jsx
        <CustomTextInput
          title="Text"
          fieldName="text"
          value={value}
          onChange={handleChange}
          hasErrors={checkFunction}
          errorMessage={errorMessage}
        />
```

apps/mull-ui/src/app/pages/login/login.tsx
```diff
@@ -0,0 +1,88 @@
...
+        />
+
+        <CustomTextInput
+          title="Password"
+          fieldName="password"
+          value={formik.values.password}
+          onChange={formik.handleChange}
+          hasErrors={formik.touched.password && !!formik.errors.password}
+          errorMessage={formik.errors.password}
+          isPassword={true}
```
bit172: Just FYI, you can just write `isPassword` and it will set it to true.

apps/mull-ui/src/app/pages/login/login.tsx
```diff
@@ -0,0 +1,88 @@
...
+          <a href="/register" className="register-login-page">
+            {' '}
+            Sign up here!
+          </a>
+        </div>
+
+        <div className="separator">or</div>
+
+        <button className="twitter">
+          <TwitterIcon />
```
bit172: These SVGs will probably change since #83 is coming but it's fine for now.

apps/mull-ui/src/app/pages/register/register.scss
```diff
@@ -0,0 +1,65 @@
...
+
+@media (min-width: 1200px) {
+  .register,
+  .custom-text-input-container {
+    width: 22%;
+  }
+  .register-mull-logo {
+    width: 10%;
+  }
+}
```
bit172: What are these new breakpoints for? Should they be added to variables?

apps/mull-ui/src/app/pages/register/register.tsx
```diff
@@ -0,0 +1,67 @@
...
+export interface RegisterProps {
+  history: History;
+}
+
+export const Register = ({ history }) => {
+  const formik = useFormik({
+    initialValues: {
+      name: '',
+      email: '',
+      password: '',
```
bit172, line 20: I know it's not on the mockup but in our backend, we currently require a date of birth for registration. This is something we need to discuss in the next meeting.

apps/mull-ui/src/app/pages/register/register.scss
```diff
@@ -0,0 +1,65 @@
...
+  color: #000000;
+  background-color: rgba(39, 176, 154, 0.6);
+  border: none;
+  height: 45px;
+  cursor: pointer;
+  border-radius: 16px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  flex-direction: row;
```
bit172: Pretty sure you don't need those for button styling.

apps/mull-ui/src/app/pages/register/register.scss
```diff
@@ -0,0 +1,65 @@
...
+.custom-text-input-container,
+.register {
+  align-self: center;
+  justify-self: center;
+}
+
+.register,
+.custom-text-input-container {
+  width: 100%;
+}
```
bit172: These can be combined.

apps/mull-ui/src/app/pages/register/register.scss
```diff
@@ -0,0 +1,65 @@
...
+.register-mull-logo {
+  padding-top: 8em;
+  display: block;
+  margin: 0 auto;
+  width: 30%;
+  padding-bottom: 3em;
+}
+
+.register-container {
+  display: grid;
```
bit172, line 12: Would it not be more simple to use a flexbox instead of a grid? I see you're centering things individually and a flexbox could help. Since every thing is in a column, I don't think grid is necessary.

apps/mull-ui/src/app/pages/login/login.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { ReactComponent as GoogleIcon } from '../../../assets/icons/login-icons/google.svg';
+import { ReactComponent as FacebookIcon } from '../../../assets/icons/login-icons/facebook.svg';
+
+import './login.scss';
+
+export interface LoginProps {
+  history: History;
+}
+
+export const Login = ({ history }) => {
```
bit172: No explicit export needed here. There's already an export default at the bottom of the file.

apps/mull-ui/src/app/pages/register/register.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { useFormik } from 'formik';
+import logo from '../../../assets/mull-logo.png';
+import './register.scss';
+
+/* eslint-disable-next-line */
+export interface RegisterProps {
+  history: History;
+}
+
+export const Register = ({ history }) => {
```
bit172: No explicit export needed here. There's already an export default at the bottom of the file.

apps/mull-ui/src/app/pages/login/login.scss
```diff
@@ -0,0 +1,136 @@
...
+}
+
+.google {
+  background-color: #4285f4;
+}
+
+.facebook {
+  background-color: #4267b2;
+  padding-left: 1.2em;
+}
```
bit172: Should these colors be stored in variables?

apps/mull-ui/src/app/pages/login/login.tsx
```diff
@@ -0,0 +1,88 @@
...
+        <button type="submit" className="login">
+          Login
+        </button>
+
+        <div className="sign-up">
+          Don't have an account?{' '}
+          <a href="/register" className="register-login-page">
+            {' '}
+            Sign up here!
+          </a>
```
bit172: Consider using a Link component from React Router.

apps/mull-ui/src/app/pages/login/login.scss
```diff
@@ -0,0 +1,136 @@
...
+.separator,
+.twitter {
+  align-self: center;
+  justify-self: center;
+}
+
+.separator {
+  display: flex;
+  align-items: center;
+  color: #27b09a;
```
bit172: Use the color variable here.

apps/mull-ui/src/app/pages/login/login.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { useFormik } from 'formik';
+import logo from '../../../assets/mull-logo.png';
+import { ReactComponent as TwitterIcon } from '../../../assets/icons/login-icons/twitter.svg';
+import { ReactComponent as GoogleIcon } from '../../../assets/icons/login-icons/google.svg';
+import { ReactComponent as FacebookIcon } from '../../../assets/icons/login-icons/facebook.svg';
+
+import './login.scss';
+
+export interface LoginProps {
+  history: History;
```
bit172, line 16: This is not the right History type. You should import just import the history package and it should be fine.

apps/mull-ui/src/app/pages/login/login.tsx
bit172: Missing prop declaration.

apps/mull-ui/src/app/pages/register/register.tsx
bit172: Missing prop declaration

bit172 (changes requested):
The page looks good, but there are quite a few things that I would like to see changed. Reach out to me if you need more clarification. Good work!

apps/mull-ui/src/app/pages/register/register.scss
cristian-aldea: don't fear the flex @JimLau49

apps/mull-ui/src/app/pages/login/login.tsx
cristian-aldea: It seems that components generated with Nx have this double export by default, although possibly redundant, I don't think it's worth changing.

apps/mull-ui/src/app/pages/login/login.scss
cristian-aldea:
```suggestion
  color: $primary-color;
```

apps/mull-ui/src/app/pages/login/login.tsx
cristian-aldea: `import { History } from 'history';`

apps/mull-ui/src/app/pages/login/login.tsx
JimLau49: oohh nice suggestion will include :)

apps/mull-ui/src/app/pages/register/register.tsx
JimLau49: done

apps/mull-ui/src/app/pages/register/register.tsx
leojrsilao: Was wondering which prop declaration that you're talking about?

cristian-aldea: He means that you're not declaring what type your props are
```suggestion
export const Register = ({ history }: RegisterProps) => {
```

apps/mull-ui/src/app/app.tsx
leojrsilao: Yeah, I think that for authentication to show routes should be done when the #30 is done. We will leave the showNavigation function for now but will replace it when the backend is done.

apps/mull-ui/src/app/pages/login/login.tsx
bit172: I think it's the export option in the VS Code extension. Just unchecking it when generating the component should be fine.

apps/mull-ui/src/app/app.tsx
leojrsilao: We'll leave it for now, but it will definitely need to be changed once the backend is done.

apps/mull-ui/src/app/app.tsx
leojrsilao: As mentioned in #30, we will implement the authentication and this will hide the pages if you're not authenticated. We will leave it as it is for now.

apps/mull-ui/src/app/pages/register/register.scss
leojrsilao: Breakpoints are for the widths to change for some inputs / buttons

apps/mull-ui/src/app/pages/register/register.scss
leojrsilao: Simply used grid because the gap property works on all browsers while the gap property on flex works only on some browsers

apps/mull-ui/src/app/app.tsx
cristian-aldea: This still hasn't been replied to. Its no big deal but I just want to make sure you saw this

bit172: The implementation for this will most likely change in #30 and would no longer require the use of the hook. IMO, I would just let it go for now.

libs/ui-lib/src/lib/custom-text-input/custom-text-input.tsx
cristian-aldea: They could do that. But I think that having them pass the type as a prop would make the custom component bahv more closely to a real `<input>` element, which would make it easier to use.

apps/mull-ui/src/app/pages/create-event/create-event.scss
```diff
@@ -9,12 +9,17 @@
...
     margin-bottom: 1.875rem;
   }
 
   .custom-time-picker-container {
     margin: 0 0 1.25rem;
   }
+
+  .date-calendar-container {
+    width: 100%;
+  }
```
cristian-aldea: This is dead css which has the same effect as
```scss
.date-calendar-container,
.custom-text-input-container {
  width: 100%;
}
```

below. Remove one or the other

apps/mull-ui/src/app/pages/login/login.scss
```diff
@@ -0,0 +1,143 @@
+@import 'variables';
+
+.login-container {
+  display: grid;
+  grid-template-columns: 1fr;
+  grid-row-gap: 1.5em;
+  padding-top: 2.5em;
+  padding-left: 2em;
+  padding-right: 2em;
```
cristian-aldea: You're using em, but I don't think the spacing here will change based on font size, you can set it to rem:
```suggestion
  grid-row-gap: 1.5rem;
  padding-top: 2.5rem;
  padding-left: 2rem;
  padding-right: 2rem;
```

apps/mull-ui/src/app/pages/login/login.scss
```diff
@@ -0,0 +1,143 @@
...
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  flex-direction: row;
+}
+
+.twitter,
+.google,
+.facebook {
+  color: #ffffff;
```
cristian-aldea:
```suggestion
  color: white;
```

apps/mull-ui/src/app/pages/register/register.scss
bit172: I guess it's fine for now. If it causes problems in the future. I can convert it into a flexbox. 🚀

apps/mull-ui/src/app/pages/login/login.scss
```diff
@@ -0,0 +1,143 @@
...
+  display: grid;
+  grid-template-columns: 1fr;
+  grid-row-gap: 1.5em;
+  padding-top: 2.5em;
+  padding-left: 2em;
+  padding-right: 2em;
+}
+
+.facebook {
+  margin-bottom: 2em;
```
cristian-aldea:
```suggestion
  margin-bottom: 2rem;
```

apps/mull-ui/src/app/pages/register/register.scss
bit172: They are still hard-coded. We should probably start adding new breakpoints to our variables so we cover a wider range of viewports. Good for now.

apps/mull-ui/src/app/pages/login/login.scss
```diff
@@ -0,0 +1,143 @@
...
+.separator::after {
+  margin-left: 0.25em;
+}
+
+.login,
+.twitter,
+.google,
+.facebook {
+  width: 100%;
+  height: 45px;
```
cristian-aldea:
```suggestion
  height: 2.8rem;
```

apps/mull-ui/src/app/pages/login/login.scss
```diff
@@ -0,0 +1,143 @@
...
+.google,
+.facebook {
+  color: #ffffff;
+  font-weight: bold;
+}
+
+.login {
+  font-weight: bold;
+  color: #000000;
+  background-color: rgba(39, 176, 154, 0.6);
```
cristian-aldea:
```suggestion
  background-color: rgba($color: $primary-color, $alpha: 0.6);
```
Or if you're feeling dangerous:
```suggestion
  background-color: rgba($primary-color, 0.6);
```

apps/mull-ui/src/app/pages/login/login.scss
```diff
@@ -0,0 +1,143 @@
...
+  border: none;
```
cristian-aldea:
```suggestion
```

apps/mull-ui/src/app/pages/login/login.scss
```diff
@@ -0,0 +1,143 @@
...
+  background-color: $google;
+}
+
+.facebook {
+  background-color: $facebook;
+  padding-left: 1.2em;
+}
+
+.login-mull-logo {
+  padding-top: 2em;
```
cristian-aldea: This element's font size is the same as the root element. No need to use em
```suggestion
  padding-top: 2rem;
```

I'll stop leaving comments about em. Just make sure to go over the code and remove instances of em if they don't fit.

You can ask me if you don't know when to use em vs. rem. Short story you most likely want to use rem.

apps/mull-ui/src/app/pages/login/login.scss
```diff
@@ -0,0 +1,143 @@
...
+.twitter,
+.google,
+.facebook {
+  color: #ffffff;
+  font-weight: bold;
+}
+
+.login {
+  font-weight: bold;
+  color: #000000;
```
cristian-aldea:
```suggestion
  color: black;
```

bit172 (dismissed):
LGTM! There are small things that I wanted to be changed but I think this can be merged and those can be fixed in their own tickets. Good work!

**Acceptance Criteria**

- [x] Documentation present for every new component and function where necessary
- [x] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=2079%3A70811)

apps/mull-ui/src/app/pages/login/login.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+        <button type="submit" className="login">
+          Login
+        </button>
+
+        <div className="sign-up">
+          Don't have an account?{' '}
+          <Link to="/register" className="register-login-page">
+            {' '}
+            Sign up here!
```
cristian-aldea: You insert two spaces in two places. I think its possible to insert both in one place
```suggestion
          Don't have an account?{'  '}
          <Link to="/register" className="register-login-page">
            Sign up here!
```

apps/mull-ui/src/app/pages/login/login.tsx
```diff
@@ -0,0 +1,91 @@
...
+          </Link>
+        </div>
+
+        <div className="separator">or</div>
+
+        <button className="twitter">
+          <FontAwesomeIcon icon={faTwitter} size="2x" />
+          &nbsp; Continue with Twitter
```
cristian-aldea, line 73: You can use {' '} instead of &nbsp; to be consistent
```suggestion
          {' '}Continue with Twitter
```
as for the buttons below

apps/mull-ui/src/app/pages/register/register.scss
```diff
@@ -0,0 +1,67 @@
+@import 'variables';
```
cristian-aldea, line 1: Same comments as for `apps/mull-ui/src/app/pages/login/login.scss`

apps/mull-ui/src/app/pages/register/register.tsx
```diff
@@ -0,0 +1,77 @@
...
+        <button type="submit" className="register">
+          Create Account
+        </button>
+
+        <div className="login-link">
+          Already have an acccount?{' '}
+          <Link to="/login" className="login-here">
+            {' '}
+            Login here!
+          </Link>
```
cristian-aldea: Same as before
```suggestion
          Already have an acccount?{'  '}
          <Link to="/login" className="login-here">
            Login here!
          </Link>
```

cristian-aldea (changes requested):
I see you left hardcoded breakpoints. I agree with @bit172 that having more breakpoints could be a seperate bug ticket. For now you should just having the width of the login and register page be the same width (35%) as the page-container

apps/mull-ui/src/app/pages/login/login.tsx
JimLau49: Tried it, this doesn't seem to work. From the images below when using {' '} it actually doesn't create a space. 

image

image

JimLau49:
> I see you left hardcoded breakpoints. I agree with @bit172 that having more breakpoints could be a seperate bug ticket. For now you should just having the width of the login and register page be the same width (35%) as the page-container

@TheGreatMarkus Putting the widths to 35% in the container, we're getting this outcome 

image

```.login-container {
  display: grid;
  grid-template-columns: 1fr;
  grid-row-gap: 1.5rem;
  padding-top: 2.5rem;
  padding-left: 2rem;
  padding-right: 2rem;
  width: 35%;
}

Maybe we're misunderstanding something though

File: apps/mull-ui/src/app/pages/login/login.tsx (hunk 35 above)
Comment (cristian-aldea) [reply apps/mull-ui/src/app/pages/login/login.tsx:73]: That's my bad, I assumed it would work without testing

Comment (cristian-aldea) [reply apps/mull-ui/src/app/pages/login/login.tsx:73]: Alternatively, you can put a margin right on the icon and put the text in a div (or maybe the div is not needed). Idk if using &nbsp; is a good idea.

File: apps/mull-ui/src/app/pages/create-event/create-event.scss [hunk 38]
```diff
@@ -15,6 +15,11 @@
   .custom-time-picker-container {
     margin: 0 0 1.25rem;
   }
+
+  .date-calendar-container,
+  .custom-text-input-container {
+    width: 100%;
+  }
```
cristian-aldea: This is a duplicate of the scss on lines 68-71. I'd suggest doing it like Ragith did in his PR: [link](https://github.com/RGPosadas/Mull/pull/77/files#diff-49e59426c9c7e504adb609db55b2111dc173b3bd305b65fe35f379d4da54ab77R19-R36)

Set 100% on each class selector individually, and make sure to not do it twice.

cristian-aldea:
> @TheGreatMarkus Putting the widths to 35% in the container, we're getting this outcome

So what's happening is that you're setting the width of the container to 35%, but you didn't remove the code you put which restricts the width of the children at certain hardcoded breakpoints. You can see this if you inspect the page:
image

If you do remove the code with the hardcoded breakpoints you get the following:
image

If you need help lemme know as we will get in a call. Essentially, just do what you did for the create event page with the `page-container` class. you have 35% width on desktop, and 100% on mobile. Don't copy the padding on that class though, since that's to compensate the fact that the navbar and header are there.

What i would do would be to create a new class in `app.scss`, maybe something called `.page-container-no-nav`, which would be the following in app.scss:

```scss

.page-container {
  padding: $header-height-mobile 0 $nav-bar-height-mobile;
}

.Toastify__toast--success {
  background: $primary_color;
}

@media (min-width: $breakpoint-mobile) {
  .Toastify__toast-container {
    top: 7.5rem;
  }
  .page-container,
  .page-container-no-nav {
    width: 35%;
    margin: auto;
  }

  .page-container {
    padding: $nav-bar-height-desktop 0 0;
  }
}

```

And then you could apply this `.page-container-no-nav` class to your register and login containers.

apps/mull-ui/src/app/pages/register/register.scss
```diff
@@ -0,0 +1,67 @@
...
+  width: 30%;
+  padding-bottom: 3rem;
+}
+
+.register-container {
+  display: grid;
+  grid-template-columns: 1fr;
+  grid-row-gap: 1.5rem;
+  margin: 0 auto;
+  width: 95%;
```
cristian-aldea:
```suggestion
  display: grid;
  grid-template-columns: 1fr;
  grid-row-gap: 1.5rem;
  padding-top: 2.5rem;
  padding-left: 2rem;
  padding-right: 2rem;
```

You applied my suggestion in the login page, but not here in the register page

apps/mull-ui/src/app/pages/register/register.scss
```diff
@@ -0,0 +1,67 @@
...
+}
+
+.register {
+  font-weight: bold;
+  color: #000000;
+  background-color: rgba(39, 176, 154, 0.6);
```
cristian-aldea:
```suggestion
  background-color: rgba($color: $primary-color, $alpha: 0.6);
```

apps/mull-ui/src/app/pages/register/register.scss
```diff
@@ -0,0 +1,67 @@
...
+  display: grid;
+  grid-template-columns: 1fr;
+  grid-row-gap: 1.5rem;
+  margin: 0 auto;
+  width: 95%;
+}
+
+.register {
+  font-weight: bold;
+  color: #000000;
```
cristian-aldea:
```suggestion
  color: white;
```

apps/mull-ui/src/app/pages/login/login.scss
```diff
@@ -0,0 +1,142 @@
...
+  .facebook,
+  .custom-text-input-container,
+  .separator {
+    width: 22%;
+  }
+
+  .login-mull-logo {
+    width: 10%;
+  }
+}
```
cristian-aldea: Please remove this code and try implementing the `.page-container-no-nav` suggestion I gave earlier.

I know that it might look a bit wide on desktop, but we can fix that by adding more breakpoints in another pr. For now just stick with what was done in the create event page

apps/mull-ui/src/app/pages/register/register.scss
```diff
@@ -0,0 +1,67 @@
...
+  background-color: rgba(39, 176, 154, 0.6);
+  border: none;
+  height: 45px;
```
cristian-aldea:
```suggestion
  height: 2.8rem;
```

apps/mull-ui/src/app/pages/register/register.scss
```diff
@@ -0,0 +1,67 @@
...
+  cursor: pointer;
+  border-radius: 16px;
+  font-size: 16px;
```
cristian-aldea: Not needed, font-size is 16px = 1rem by default

```suggestion
```

apps/mull-ui/src/app/pages/login/login.scss
```diff
@@ -0,0 +1,142 @@
...
+  padding-bottom: 3rem;
+}
+
+.sign-up {
+  text-align: center;
+  font-size: 14px;
+}
+
+.register-login-page {
+  font-size: 14px;
```
cristian-aldea: Might be best to define these in terms of the root font size, in case we change that later:
```suggestion
  font-size: 0.85rem;
}

.register-login-page {
  font-size: 0.85rem;
```

apps/mull-ui/src/app/pages/register/register.scss
```diff
@@ -0,0 +1,67 @@
...
+
+.custom-text-input-container,
+.register {
+  align-self: center;
+  justify-self: center;
+  width: 100%;
+}
+
+.login-here {
+  font-size: 14px;
```
cristian-aldea: Same as for login:
```suggestion
  font-size: 0.85rem;
```

apps/mull-ui/src/app/pages/register/register.scss
```diff
@@ -0,0 +1,67 @@
...
+
+@media (min-width: 1200px) {
+  .register,
+  .custom-text-input-container {
+    width: 22%;
+  }
+  .register-mull-logo {
+    width: 10%;
+  }
+}
```
cristian-aldea: Same as for login, remove this and have the `.page-container-no-nav` suggestion.

cristian-aldea (commented):
Sorry for the double review whammy, A bunch of stuff that was fixed in login wasn't in register which lead to inconsistencies. Also, saw a few other problems I didn't notice before.

apps/mull-ui/src/app/pages/register/register.scss
JimLau49: Oops my bad

apps/mull-ui/src/app/pages/create-event/create-event.scss
cristian-aldea: @r-saba  already applied this styling in his PR, you can remove it when you update your branch with his changes

apps/mull-ui/src/app/pages/login/login.tsx
```diff
@@ -0,0 +1,92 @@
...
+      password: Yup.string().required('Password is required.'),
+    }),
+
+    onSubmit: (values) => {},
+  });
+
+  return (
+    <form onSubmit={formik.handleSubmit}>
+      <div className="page-container-no-nav">
+        <div className="login-container">
```
cristian-aldea: No need to have an extra div, you should be able to apply the styling to the same div.

apps/mull-ui/src/app/pages/register/register.tsx
```diff
@@ -0,0 +1,78 @@
...
+      password: Yup.string().required('Password is required.'),
+    }),
+
+    onSubmit: (values) => {},
+  });
+
+  return (
+    <form onSubmit={formik.handleSubmit}>
+      <div className="page-container-no-nav">
+        <div className="register-container">
```
cristian-aldea: No need to have an extra div, you should be able to apply the styling to the same div.

cristian-aldea (dismissed):
Thank you so much for addressing my comments this time. I saw a few more issues but they don't really matter or affect anything, so I won't block it. Good work!

- [x] Documentation present for every new component and function where necessary
- [x] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=2079%3A70811)

apps/mull-ui/src/app/pages/login/login.tsx
JimLau49: oh yeah oops, was testing something forgot to merge these together, ill do this then rebase

cristian-aldea (approved):

bit172 (approved):

Issue #73, opened 2020-11-11
TASK-19: Multiple Registration Options Back-end


r-saba:
**Task Description**

As a developer, I want to develop the business logic for multiple registration options to be able to implement the user flow.

**Acceptance Criteria**
- [ ] Registration is possible using an existing service (Google, Twitter, Facebook) or email.
- [ ] Documentation present for every new component and function where necessary
- [ ] Unit tests are implemented.

PR #90, opened 2020-11-22
TASK-19: Multiple Registration Options Back-end

bit172:
This PR implements the backend of Multiple Registration Options #73.

**Steps to Test:**
1. Go to the [slack](https://bug-bytes.slack.com/archives/CS5P4NL5Q/p1606083803001700) and update your ```.env``` file.
1. Install dependencies using ```npm i```.
1. Start the server using ```nx serve mull-api```.
1. In a browser, go to ```localhost:3333/api/auth/google``` , ```localhost:3333/api/auth/facebook```, ```localhost:3333/api/auth/twitter```
1. Follow the instructions for each service.
1. You should get a returned user after a successful OAuth validation.

**Notes:**
The accounts on the slack link can be used to test if you're not comfortable using your personal accounts.
There is a chance that Twitter OAuth might not work. If that is the case, reach out to me on slack.

Facebook requires me to add anyone who wants to test as a developer on the app. I will try inviting everyone but if it doesn't work, try logging in to the Facebook Developer account and adding yourself as a developer. (This is caused by the app being in development mode)


apps/mull-api/src/app/app.module.ts
```diff
@@ -47,6 +50,14 @@ import { environment } from '../environments/environment';
...
       },
     }),
+    SessionModule.forRoot({
+      session: {
+        name: 'mull-app-session',
+        secret: environment.session.secret,
+        resave: false,
+        saveUninitialized: false,
+      },
+    }),
```
r-saba, line 60: Do we plan to do anything with sessions in the future or is it just for Twitter OAuth?

apps/mull-api/src/app/auth/auth.service.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (!req.user) return null;
+    return { user: req.user };
+  }
+
+  async validateOAuthUser(
+    user: Partial<User>,
+    callback: (error: string, user?: Partial<User>, info?: string) => void
+  ) {
+    const users: User[] = await this.userService.findByEmail(user.email);
+    if (users && users.length === 1 && user.type === users[0].type) {
```
r-saba: Can you explain this line? Also why do we have to check that the type is the same?

apps/mull-api/src/app/auth/auth.service.ts
```diff
@@ -0,0 +1,42 @@
...
+    } else {
+      const newUser = await this.userService.create(user as CreateUserInput);
+      callback(null, newUser);
+    }
+  }
+
+  async validateUser(email: string, pass: string): Promise<Partial<User>> {
+    const users: User[] = await this.userService.findByEmail(email);
+    if (users && users.length === 1) {
+      const same = await compare(pass, users[0].password);
```
r-saba: Rename `same` to something more descriptive. `validPassword`?

apps/mull-api/src/app/entities/user.entity.ts
```diff
@@ -22,7 +22,7 @@ export class User implements IUser {
   id: number;
 
   @Field()
-  @Column()
+  @Column({ nullable: true })
```
r-saba, line 25: Why is null allowed for passwords?

apps/mull-api/src/app/entities/user.entity.ts
```diff
@@ -44,9 +44,16 @@ export class User implements IUser {
...
   dob: Date;
 
+  @Field(() => Int)
+  @Column({
+    type: 'enum',
+    enum: UserType,
+  })
+  type: UserType;
```
r-saba: Is this for keeping track of which method a user used to login (local, google, facebook, twitter)?

apps/mull-api/src/app/user/user.service.ts
```diff
@@ -19,16 +21,23 @@ export class UserService {
...
-    const newUser = this.userRepository.create({ name, dob, email, password });
-    await this.userRepository.save(newUser);
-    return newUser;
+    if (userInput.type === UserType.LOCAL) {
+      const salt = await genSalt(10);
+      const hashed = await hash(userInput.password, salt);
+      userInput.password = hashed;
+    }
+    const newUser = this.userRepository.create(userInput);
+    return await this.userRepository.save(newUser);
```
r-saba: I haven't tested but can you check if we can do this in one line like how we usually do?
```suggestion
    return await this.userRepository.save(userInput);
```

r-saba (changes requested):
This is such a good PR, even your tests are done so well!

How can I test the local auth?

I get this error when I try to login with my personal Facebook account. Can you take a look? I sent you the stack trace as text on slack to make it easier.
image

apps/mull-api/src/app/app.module.ts
bit172: Twitter login is using OAuth 1.0 which requires sessions, so it's probably going to be used only for that but it can be useful to track the users in the backend if we need to.

apps/mull-api/src/app/entities/user.entity.ts
bit172: Yes, I know a better name for the type and the column can be used so if you come up with one then send it my way 😅.

apps/mull-api/src/app/entities/user.entity.ts
bit172: OAuth accounts don't have passwords. I'm not sure what the best approach would be but I made passwords optional for now.

bit172:
> This is such a good PR, even your tests are done so well!
> 
> How can I test the local auth?
> 
> I get this error when I try to login with my personal Facebook account. Can you take a look? I sent you the stack trace as text on slack to make it easier.
> image

I think it has something to do with the scope parameter for the strategy. The type provided by the ```@types/passport-facebook``` library doesn't seem to have that option. I will take a look into this issue.

For local auth testing, I just used the GraphQL playground to create a user, since this ticket does not include login. This would be done in #30.

apps/mull-api/src/app/user/user.service.ts
bit172: I get this error when I try to write it that way. 
image
I also remember that I wrote it that way since I ran into the same problem.

apps/mull-api/src/app/auth/auth.service.ts
bit172: I think I should change the logic a bit, but it checks if a user with the email exists and allows multiple accounts to be created with the same email depending on the service that is being used. Not sure if this is something we want. (multiple accounts with the same email).

apps/mull-api/src/app/entities/user.entity.ts
RGPosadas: RegistrationMethod is an idea

RGPosadas (changes requested):
I tested on local as well as Google and Twitter OAuth, and all three worked well for me.
As per Ragith's comment, I also get the same error he posted when trying to register using Facebook.
Otherwise, good work 👍

apps/mull-api/src/app/user/user.service.ts
r-saba: I was able to do it in one like so. Tested it and it works.
```suggestion
    return await this.userRepository.save({...userInput});
```

apps/mull-api/src/app/app.module.ts
r-saba: Since we have sessions already we can use it to manage authenticated state of the user in the login ticket. NestJS docs suggests [JWT or express sessions](https://docs.nestjs.com/security/authentication)

apps/mull-api/src/app/entities/user.entity.ts
r-saba: I think this is fine then.

apps/mull-api/src/app/user/user.service.ts
```diff
@@ -19,16 +21,23 @@ export class UserService {
...
     const { friends } = await this.userRepository.findOne(id, { relations: ['friends'] });
     return friends;
   }
 
   async create(userInput: CreateUserInput): Promise<User> {
-    const { dob, name, email, password } = userInput;
-    const newUser = this.userRepository.create({ name, dob, email, password });
-    await this.userRepository.save(newUser);
-    return newUser;
+    if (userInput.type === UserType.LOCAL) {
```
RGPosadas: Sorry, missed this in my initial review.
I was able to register using an empty string as a password, so we should add a check for that here as well.

RGPosadas: Or hmm the frontend implementation can (and should) also do the form validation, so I am unsure of my comment now.

bit172: I could add backend validation. I think we should start having something for it.

cristian-aldea: The more validation the better

apps/mull-api/src/app/auth/auth.service.ts
RGPosadas: As per our post-scrum, we keep this as is as this is what we observe in other websites as well.

apps/mull-api/src/app/user/user.service.ts
bit172: Decided to use the class-validator package for validation as it was recommended by the nestjs docs.

apps/mull-api/src/app/auth/auth.controller.ts
```diff
@@ -0,0 +1,45 @@
...
+import { AuthGuard } from '@nestjs/passport';
+
+@Controller('auth')
+export class AuthController {
+  constructor(private readonly authService: AuthService) {}
+
+  @Get('google')
+  @UseGuards(AuthGuard('google'))
+  async googleAuth() {
+    // noop
```
cristian-aldea, line 13: Wait so what is happening exactly when I go to `localhost:333/api/auth/(facebook|google|twitter)`?

Does it have to do with the guard?

apps/mull-api/src/app/auth/local.strategy.ts
```diff
@@ -0,0 +1,20 @@
+import { Strategy } from 'passport-local';
+import { PassportStrategy } from '@nestjs/passport';
+import { Injectable, UnauthorizedException } from '@nestjs/common';
+import { AuthService } from './auth.service';
+import { User } from '../entities/user.entity';
+
+@Injectable()
+export class LocalStrategy extends PassportStrategy(Strategy) {
```
cristian-aldea, line 8: Is there a way to try out the local strategy in this PR?

apps/mull-api/src/app/auth/facebook.strategy.ts
```diff
@@ -0,0 +1,37 @@
...
+import { AuthService } from './auth.service';
+import { RegistrationMethod } from '@mull/types';
+
+@Injectable()
+export class FacebookStrategy extends PassportStrategy(Strategy, 'facebook') {
+  constructor(private authService: AuthService) {
+    super({
+      clientID: environment.auth.facebook.clientId,
+      clientSecret: environment.auth.facebook.clientSecret,
+      callbackURL: 'http://localhost:3333/api/auth/facebook/redirect',
```
cristian-aldea, line 15: How is this gonna work when the application is deployed? Might need to put this in the environment.ts file

apps/mull-api/src/main.ts
```diff
@@ -13,6 +13,7 @@ async function bootstrap() {
   const globalPrefix = 'api';
   app.enableCors();
   app.setGlobalPrefix(globalPrefix);
+  app.useGlobalPipes(new ValidationPipe());
```
cristian-aldea, line 16: What does this do?

cristian-aldea (changes requested):
After adding myself as a developer on the Mull app on Facebook, I can say that all OAuth methods worked great and returns a user successfully!

I didn't run into the issue that Ritchelle and Ragith ran into, so I'll consider it fixed.

- [ ] Registration is possible using an existing service (Google, Twitter, Facebook) or email.
- [x] Unit tests are implemented.

However, your instructions didn't mention how to login using the local strategy. I see that its implemented, but I'm not sure how to test it and create a new user

My comments are mostly questions, since I wasn't very involved in the backend implementation so far.

JimLau49:
Heyo, gonna be taking a look into this after working hours today, as im assigned to the main story, it would do me well to understand how this works so that we can connect TASK-18 to this easily.

apps/mull-api/src/app/auth/facebook.strategy.ts
bit172: Right, I could add a BASE_URL variable

apps/mull-api/src/app/auth/auth.controller.ts
bit172: Yes, the Guard will handle it. I needed to create a function so that I could have a REST endpoint.

apps/mull-api/src/main.ts
bit172: This is to enable validation on all endpoints (including GraphQL). I followed instructions [here](https://docs.nestjs.com/techniques/validation) to get started and this works in conjunction with the class-validator package.

apps/mull-api/src/app/auth/local.strategy.ts
bit172: I think the goal was to have this be done in #30. Technically, this ticket was only considering registration, but for OAuth methods I included logging in since it felt like it was blocking the flow in the backend.

apps/mull-api/src/app/auth/auth.controller.ts
bit172: The Guard essentially runs the validate method of the different Strategies and sets the user variable in the HTTP request.

apps/mull-api/src/app/auth/facebook.strategy.ts
cristian-aldea: this can be done in #30, just wanted to mention it

apps/mull-api/src/app/auth/local.strategy.ts
cristian-aldea: the local strategy is part of the ac for task 19. I don't mind doing it in #30 thou

cristian-aldea (dismissed):
- [x] Registration is possible using an existing service (Google, Twitter, Facebook) or email.
- [x] Unit tests are implemented.

RGPosadas (approved):
* I can no longer create a user with an empty string, so thank you for fixing that!
* I can now confirm that Facebook works as well.

LGTM! 🚀 

**Acceptance Criteria**
- [x] Registration is possible using an existing service (Google, Twitter, Facebook) or email.
- [x] Documentation present for every new component and function where necessary
- [x] Unit tests are implemented.

apps/mull-api/src/app/user/user.service.ts
RGPosadas: I can no longer create a user with an empty string as password, therefore marking this as resolved.

r-saba (approved):
LGTM! I can register with Facebook now and my other comments were addressed. Thanks!

**Acceptance Criteria**
- [x] Registration is possible using an existing service (Google, Twitter, Facebook) or email.
- [x] Documentation present for every new component and function where necessary
- [x] Unit tests are implemented.

apps/mull-api/src/app/auth/auth.service.ts
```diff
@@ -0,0 +1,49 @@
...
+    }
+    if (!hasSameRegistrationMethod) {
+      const newUser = await this.userService.create(user as CreateUserInput);
+      callback(null, newUser);
+    }
+  }
+
+  async validateUser(email: string, pass: string): Promise<Partial<User>> {
+    const users: User[] = await this.userService.findByEmail(email);
+    if (users && users.length === 1) {
```
JimLau49, line 40: Nice change, I remember that this if statement was more verbose than i liked, but didnt have time to comment.

JimLau49 (approved):
Really nice work Jonathan 🚀 looking forward to working on connecting the frontend to this! Testing wise you definitely went very thorough and it shows.

**Acceptance Criteria**
- [x] Registration is possible using an existing service (Google, Twitter, Facebook) or email.
- [x] Documentation present for every new component and function where necessary
- [x] Unit tests are implemented.

Issue #62, opened 2020-11-08
TASK-14: Event Card


cristian-aldea:
**Task Description**

As a developer, I want to implement a generic event card component so that it can be available across the application to support other user stories.

**Acceptance Criteria**
- [ ] Documentation present for relevant new components and functions
- [ ] Unit and UI tests are implemented for every component and function
- [ ] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=2000%3A343283)

r-saba:
Closing this task since #65 was pushed to backlog

PR #94, opened 2020-11-27
TASK-14: Event Card

cristian-aldea:
This PR addressed #62 

You can see the event card by going to `http://localhost:4200/test-event-card`

Notes:
* The join and share buttons are clickable, but they don't do anything for now.
* The number of the date isn't bold since I felt it didn't look right, we can discuss whether the mockup or my implementation needs to be updated.
* There are a few placeholders in this components for now, since much of the implementation it relies on (users joining events, friends, share, etc) isn't implemented yet.
* created new `formatDate` function, which is currently used in event card and event header.

### DIFF
--- a/apps/mull-ui/src/app/app.tsx
+++ b/apps/mull-ui/src/app/app.tsx
@@ -37,17 +37,15 @@ export const App = () => {
         <Route exact path={ROUTES.CREATE_EVENT} component={CreateEventPage} />
         {/* Temporary, to be removed */}
         <Route exact path={'/test-event-page'}>
-          <div className="page-container">
+          <div className="page-container no-padding">
             <EventPage event={dummyEvent} prevPage={'Review'} />
           </div>
         </Route>
         {/* Temporary, to be removed */}
         <Route exact path={'/test-event-card'}>
           <div className="page-container">
-            <div style={{ padding: '1rem' }}>
-              <EventCard event={dummyEvent} style={{ marginBottom: '1rem' }} />
-              <EventCard event={dummyEvent} style={{ marginBottom: '1rem' }} />
-            </div>
+            <EventCard event={dummyEvent} style={{ marginBottom: '1rem' }} />
+            <EventCard event={dummyEvent} style={{ marginBottom: '1rem' }} />
           </div>
         </Route>
         <Route exact path={ROUTES.LOGIN} component={LoginPage} />
--- a/apps/mull-ui/src/app/app.tsx
+++ b/apps/mull-ui/src/app/app.tsx
@@ -37,9 +37,7 @@ export const App = () => {
         <Route exact path={ROUTES.CREATE_EVENT} component={CreateEventPage} />
         {/* Temporary, to be removed */}
         <Route exact path={'/test-event-page'}>
-          <div className="page-container no-padding">
-            <EventPage event={dummyEvent} prevPage={'Review'} />
-          </div>
+          <EventPage event={dummyEvent} prevPage={'Review'} />
         </Route>
         {/* Temporary, to be removed */}
         <Route exact path={'/test-event-card'}>
--- a/apps/mull-ui/src/app/app.tsx
+++ b/apps/mull-ui/src/app/app.tsx
@@ -3,7 +3,7 @@ import { ToastContainer, toast } from 'react-toastify';
 import { Redirect, Route, Switch, useLocation } from 'react-router-dom';
 
 import CreateEventPage from './pages/create-event/create-event';
-import { NavigationBar, Header } from './components';
+import { NavigationBar, Header, EventCard } from './components';
 import LoginPage from './pages/login/login';
 import RegisterPage from './pages/register/register';
 
@@ -39,6 +39,15 @@ export const App = () => {
         <Route exact path={'/test-event-page'}>
           <div className="page-container">
             <EventPage event={dummyEvent} prevPage={'Review'} />
+          </div>
+        </Route>
+        {/* Temporary, to be removed */}
+        <Route exact path={'/test-event-card'}>
+          <div className="page-container">
+            <div style={{ padding: '1rem' }}>
+              <EventCard event={dummyEvent} style={{ marginBottom: '1rem' }} />
+              <EventCard event={dummyEvent} style={{ marginBottom: '1rem' }} />
+            </div>
           </div>
         </Route>
         <Route exact path={ROUTES.LOGIN} component={LoginPage} />
--- a/apps/mull-ui/src/app/components/event-card/event-card.tsx
+++ b/apps/mull-ui/src/app/components/event-card/event-card.tsx
@@ -1,0 +1,61 @@
+import { IEvent } from '@mull/types';
+import React, { useState } from 'react';
+
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faCircle, faShareAlt, faCheck, faSignInAlt } from '@fortawesome/free-solid-svg-icons';
+
+import './event-card.scss';
+
+/* eslint-disable-next-line */
+export interface EventCardProps {
+  event: Partial<IEvent>;
+  style?: React.CSSProperties;
+}
+
+export const EventCard = ({ event, style = {} }: EventCardProps) => {
+  const [joined, setJoined] = useState<boolean>(false);
+
+  const dateToString = (date: Date): string => {
+    const dateString = Intl.DateTimeFormat('en-us', {
+      month: 'short',
+      day: 'numeric',
+    }).format(date);
+    const timeString = Intl.DateTimeFormat('en-us', {
+      hour: '2-digit',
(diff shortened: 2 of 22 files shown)


apps/mull-ui/src/app/app.tsx
```diff
@@ -41,6 +41,15 @@ export const App = () => {
             <EventPage event={dummyEvent} prevPage={'Review'} />
           </div>
         </Route>
+        {/* Temporary, to be removed */}
+        <Route exact path={'/test-event-card'}>
+          <div className="page-container">
+            <div style={{ padding: '1rem' }}>
```
r-saba: This padding should be part of `page-container` so all our pages have uniform padding and less repetition in our code.

apps/mull-ui/src/app/components/event-card/event-card.tsx
```diff
@@ -0,0 +1,58 @@
...
+import React, { useState } from 'react';
+
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faShareAlt, faCheck, faSignInAlt } from '@fortawesome/free-solid-svg-icons';
+
+import { formatDate } from '../../../utilities';
+
+import './event-card.scss';
+
+/* eslint-disable-next-line */
```
r-saba: We can remove this

apps/mull-ui/src/app/components/event-card/event-card.tsx
```diff
@@ -0,0 +1,58 @@
...
+        onClick={() => setJoined(!joined)}
+        className={`event-card-join ${joined ? 'joined' : ''}`}
+      >
+        <FontAwesomeIcon icon={joined ? faCheck : faSignInAlt} />
+      </button>
+      <div className="event-card-description">
+        <div className="event-card-text">
+          <div className="event-card-title">{event.title}</div>
+          {/* TODO: Replace placeholder distance*/}
+          <div className="event-card-location">14km • {event.location.point}</div>
```
r-saba: Shouldn't the distance also be dynamic?

apps/mull-ui/src/app/components/event-card/event-card.tsx
```diff
@@ -0,0 +1,58 @@
...
+        </div>
+        {/* TODO: Replace placeholder */}
+        <div className="event-card-friends">friends</div>
```
r-saba, line 52: How come the styling for the friends part of the card is not implemented?

r-saba (commented):
No blockers just some questions I would like answered

apps/mull-ui/src/app/app.tsx
cristian-aldea: This might work for something like the create event page, but it wouldn't work for the event page, for example. I do agree that styling like this should be repeated as little as possible but idk how

apps/mull-ui/src/app/components/event-card/event-card.tsx
cristian-aldea: it is , but currently I have no way of calculating it. I left it a placeholder for now

apps/mull-ui/src/app/components/event-card/event-card.tsx
cristian-aldea: I feel like that would be a separate component, and it's used in a few places in the application. I could create a component for it, but the images would all be placeholders

apps/mull-ui/src/app/components/event-card/event-card.tsx
cristian-aldea: I'll make distance a variable which is interpolated in. And then leave a comment mentionning that it should be implemented.

r-saba: We could do what bootstrap does. They have a generic `container` class with default padding. Then you can use `container fluid` if you'd like the container to take up the full screen.

apps/mull-ui/src/app/components/event-card/event-card.tsx
r-saba: I only saw it in event-card if it's in other places it could be a separate task. If not we'd want it to be done in this

apps/mull-ui/src/app/app.tsx
r-saba: We could do what bootstrap does. They have a generic `container` class with default padding. Then you can use `container fluid` if you'd like the container to take up the full screen.

apps/mull-ui/src/app/components/event-card/event-card.tsx
r-saba: True, it's fine like this for now.

apps/mull-ui/src/app/app.tsx
cristian-aldea: Sure, I can do that real quick

apps/mull-ui/src/app/components/event-card/event-card.tsx
cristian-aldea: The event page is also meant to have it, currently, the icon that's there was more meant as a placeholder
image

r-saba: oh okay yeah let's have it as a separate task then, I will make it now

apps/mull-ui/src/app/components/event-page-header/event-page-header.tsx
```diff
@@ -9,26 +9,16 @@ import './event-page-header.scss';
...
-    }).format(date);
-
-    return `${dateString}\n${timeString}`;
-  };
-
+  const { day: startDay, month: startMonth, time: startTime } = formatDate(event.startDate);
+  const { day: endDay, month: endMonth, time: endTime } = formatDate(event.endDate);
   return (
     <div className="event-page-header">
       <MullBackButton>{prevPage}</MullBackButton>
```
bit172, line 24: This is not related to your ticket but I just noticed that pressing the back button reloads the website when going back instead of going to the previous page like a single page application. Would be worth creating a bug ticket for it.

bit172 (dismissed):
LGTM! It looks really nice! Good work! 🚀

- [x] Documentation present for relevant new components and functions
- [x] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=2000%3A343283)

apps/mull-ui/src/app/components/event-card/event-card.tsx
cristian-aldea: thanks ragith

apps/mull-ui/src/app/components/event-page-header/event-page-header.tsx
cristian-aldea: Good point, will create one

cristian-aldea:
@r-saba I addressed your comment about moving common padding to page-container. This affected the styling of all pages which use page-container:
- create event
- login
- register
- test event page
- test event card

So please quickly check these pages on mobile and desktop to make sure I didn't break them.

Have a look at https://github.com/RGPosadas/Mull/pull/94/commits/bb8a2d04ffa42066d3c1fe54cc06b61e5a5e7ccd and https://github.com/RGPosadas/Mull/pull/94/commits/8db106775d1cd64935f018fe871492f8fdecadb2 for details or my changes

@bit172 Sorry for dismissing your review, I created a bug related to what you brought up.

RGPosadas (dismissed):
Checked out the rest of our pages (login, register, create-event, test-event-page, test-event-card) and they are not broken.

I also like the subtle changes you made compared to the mockup, I find that the event card looks a bit cleaner now. So good job overall! Looks good to me 🚀 

**Acceptance Crtieria**
- [x] Documentation present for relevant new components and functions
- [x]  Unit and UI tests are implemented for every component and function
- [x] Must match Mockup

r-saba (changes requested):
Noticed there's barely a margin between the `event-card-text` and `event-card-friends` it's noticeable on smaller screen sizes when the text wraps. 
image
Here's what the white-space looks like on the mockup
image

cristian-aldea:
> Here's what the white-space looks like on the mockup

I agree that adding some more margin can work. Initially, I intentionally made the margin small since wrapping would make the whitespace seem much than it needed to be.

I'm not going to add as much margin as you see on the mockup, since that's obviously due to the mockup event card being much longer than the implementation event card in the comparison you gave.

And also the mockup appears a bit off since there's very little space between the friend's icon and the share button. That might need to be fixed.

r-saba (approved):
Thanks for work, it looks really good! LGTM

**Acceptance Criteria**
- [X] Documentation present for relevant new components and functions
- [X] Unit and UI tests are implemented for every component and function
- [X] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=2000%3A343283)

bit172 (approved):

Issue #106, opened 2020-11-30
BUG-7: SwipeableRoutes doesn't work when redirect route is present


cristian-aldea:
**Describe The Bug**

Currently in our code we have the following in `app.tsx`:
```tsx
<SwipeableRoutes>
<Route path={ROUTES.DISCOVER} component={() => <div>DISCOVER!</div>} />
<Route path={ROUTES.UPCOMING} component={() => <div>UPCOMING!</div>} />
<Route path={ROUTES.MY_EVENTS} component={() => <div>MY_EVENTS!</div>} />
<Route path={ROUTES.HOME} render={() => <Redirect to={ROUTES.DISCOVER} />} />
</SwipeableRoutes>
```

Having the redirect (the forth and last Route) causes the page to not show up under certain conditions:
![image](https://user-images.githubusercontent.com/23544999/100638229-c3545580-3301-11eb-9b08-6b59ffd4182d.png)

**Bug Strategy**
- [ ] Fix it now
- [x] Fix it later

**To Reproduce**

1. Serve the application
1. Click on the home button if you're not already on the home page
1. See error

**Expected Behavior**

The associated with the current route should be shown:
![image](https://user-images.githubusercontent.com/23544999/100638345-eaab2280-3301-11eb-8a25-cb0ebc54ce9d.png)

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional Context**

PR #108, opened 2020-12-02
BUG-7: SwipeableRoutes Not working

r-saba:
Fix for #106 

Added an exact match for the `home` route so that it is not matched on routes like `/home/discover`.


cristian-aldea (approved):

osahtout (approved):

JimLau49 (approved):
LGTM 🚀

Issue #99, opened 2020-11-29
US-7.6: Login as a User


r-saba:
**User Story Description**

As a User, I want to login so that I can use the Mull application.

**User Acceptance Flow**

1. User opens the app
1. User sees multiple ways of logging in
    a. Through an existing service (Google, Twitter, Facebook, etc)
    b. Using user's email
1. User logs in
1. User sees the home page

**Acceptance Criteria**

- [ ] User is authenticated
- [ ] User cannot access the application without logging in
- [ ] User Acceptance Flow is respected
- [ ] Documentation present where necessary
- [ ] Unit and UI tests are implemented for every component and function
- [ ] There must exist an automated Cypress acceptance test which tests the user acceptance flow
- [ ] Must match [Mockup](create-mockup-link)

PR #116, opened 2020-12-11
TASK-32: Joining an Event backend

### LINKED ISSUE #115: TASK-32: Joining an event Back-end
**Task Description**

As a developer, I want to have a mutation which will add an existing user as a participant to an event

**Acceptance Criteria**

r-saba:
This PR implements #115

To test, use the `addParticipantToEvent` mutation in any GraphQL IDE of your preference.
I suggest using [Apollo Studio Explorer](https://studio.apollographql.com/).

### DIFF
--- a/apps/mull-api/src/app/entities/user.entity.ts
+++ b/apps/mull-api/src/app/entities/user.entity.ts
@@ -21,6 +21,10 @@ registerEnumType(RegistrationMethod, {
 @Entity()
 @ObjectType()
 export class User implements IUser {
+  constructor(id?: number) {
+    this.id = id;
+  }
+
   @Field(/* istanbul ignore next */ () => ID)
   @PrimaryGeneratedColumn()
   id: number;
--- a/apps/mull-api/src/app/event/event.resolver.spec.ts
+++ b/apps/mull-api/src/app/event/event.resolver.spec.ts
@@ -12,7 +12,7 @@ const mockEventService = () => ({
   update: jest.fn((mockEventData: UpdateEventInput) => ({ ...mockEventData })),
   delete: jest.fn((id: number) => mockAllEvents.find((event) => event.id === id)),
   addParticipant: jest.fn((eventId: number, userId: number) => {
-    let event = mockAllEvents.find((event) => (event.id = eventId));
+    const event = mockAllEvents.find((event) => (event.id = eventId));
     event.participants.push(new User(userId));
     return event;
   }),
--- a/apps/mull-api/src/app/event/event.resolver.spec.ts
+++ b/apps/mull-api/src/app/event/event.resolver.spec.ts
@@ -3,6 +3,7 @@ import { EventResolver } from './event.resolver';
 import { EventService } from './event.service';
 import { CreateEventInput, UpdateEventInput } from './inputs/event.input';
 import { mockPartialEvent, mockAllEvents } from './event.mockdata';
+import { User } from '../entities';
 
 const mockEventService = () => ({
   create: jest.fn((mockEventData: CreateEventInput) => ({ ...mockEventData })),
@@ -10,6 +11,11 @@ const mockEventService = () => ({
   findAll: jest.fn(() => mockAllEvents),
   update: jest.fn((mockEventData: UpdateEventInput) => ({ ...mockEventData })),
   delete: jest.fn((id: number) => mockAllEvents.find((event) => event.id === id)),
+  addParticipant: jest.fn((eventId: number, userId: number) => {
+    let event = mockAllEvents.find((event) => (event.id = eventId));
+    event.participants.push(new User(userId));
+    return event;
+  }),
 });
 
 describe('UserResolver', () => {
@@ -46,6 +52,11 @@ describe('UserResolver', () => {
   it('should return the deleted event', async () => {
     const deletedEvent = await resolver.deleteEvent(35);
     expect(deletedEvent).toEqual(mockAllEvents.find((event) => event.id === 35));
+  });
+
+  it('should add the participant to the event', async () => {
+    const event = await resolver.addParticipantToEvent(35, 1);
+    expect(event.participants[0].id).toEqual(1);
   });
 
   it('should return the event with given id', async () => {
--- a/apps/mull-api/src/app/event/event.resolver.spec.ts
+++ b/apps/mull-api/src/app/event/event.resolver.spec.ts
@@ -55,8 +55,8 @@ describe('UserResolver', () => {
   });
 
   it('should add the participant to the event', async () => {
-    const event = await resolver.addParticipantToEvent(35, 1);
-    expect(event.participants[0].id).toEqual(1);
+    const success = await resolver.addParticipantToEvent(35, 1);
+    expect(success).toEqual(true);
   });
 
   it('should return the event with given id', async () => {
--- a/apps/mull-api/src/app/event/event.resolver.ts
+++ b/apps/mull-api/src/app/event/event.resolver.ts
@@ -32,8 +32,9 @@ export class EventResolver {
     return this.eventService.delete(id);
   }
 
-  @Mutation(() => Event)
+  @Mutation(() => Boolean)
   async addParticipantToEvent(@Args('eventId') eventId: number, @Args('userId') userId: number) {
-    return this.eventService.addParticipant(eventId, userId);
+    this.eventService.addParticipant(eventId, userId);
+    return true;
   }
 }
--- a/apps/mull-api/src/app/event/event.resolver.ts
+++ b/apps/mull-api/src/app/event/event.resolver.ts
@@ -31,4 +31,9 @@ export class EventResolver {
   async deleteEvent(@Args('id', { type: /* istanbul ignore next */ () => Int }) id: number) {
     return this.eventService.delete(id);
   }
+
+  @Mutation(() => Event)
+  async addParticipantToEvent(@Args('eventId') eventId: number, @Args('userId') userId: number) {
+    return this.eventService.addParticipant(eventId, userId);
+  }
 }
--- a/apps/mull-api/src/app/event/event.service.ts
+++ b/apps/mull-api/src/app/event/event.service.ts
@@ -1,4 +1,4 @@
-import { Injectable, UnauthorizedException } from '@nestjs/common';
+import { Injectable } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
 import { Repository } from 'typeorm';
 import { Event, User } from '../entities';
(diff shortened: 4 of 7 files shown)


apps/mull-api/src/app/event/event.resolver.spec.ts
```diff
@@ -3,13 +3,19 @@ import { EventResolver } from './event.resolver';
...
 
 const mockEventService = () => ({
   create: jest.fn((mockEventData: CreateEventInput) => ({ ...mockEventData })),
   findOne: jest.fn((id: number) => mockAllEvents.find((event) => event.id === id)),
   findAll: jest.fn(() => mockAllEvents),
   update: jest.fn((mockEventData: UpdateEventInput) => ({ ...mockEventData })),
   delete: jest.fn((id: number) => mockAllEvents.find((event) => event.id === id)),
+  addParticipant: jest.fn((eventId: number, userId: number) => {
+    const event = mockAllEvents.find((event) => (event.id = eventId));
+    event.participants.push(new User(userId));
```
ys-lin, line 16: Just a question, does `(new User(userId))` create a new user in the database? Or is it like a `find_or_create` ?

apps/mull-api/src/app/event/event.service.ts
```diff
@@ -27,6 +27,13 @@ export class EventService {
     return this.findOne(eventInput.id);
   }
 
+  async addParticipant(eventId: number, userId: number) {
+    const user = new User(userId);
```
ys-lin, line 31: Just a question, does `new User(userId)` create a new user in the DB? or is it like a `find_or_create`?

apps/mull-api/src/app/event/event.service.ts
```diff
@@ -27,6 +27,13 @@ export class EventService {
...
+    const event = await this.eventRepository.findOne(eventId, { relations: ['participants'] });
+    event.participants.push(user);
```
ys-lin, line 33: Another question, what happens if the user id already exists in the array?

apps/mull-api/src/app/event/event.resolver.spec.ts
```diff
@@ -3,13 +3,19 @@ import { EventResolver } from './event.resolver';
...
 import { mockPartialEvent, mockAllEvents } from './event.mockdata';
+import { User } from '../entities';
 
 const mockEventService = () => ({
   create: jest.fn((mockEventData: CreateEventInput) => ({ ...mockEventData })),
   findOne: jest.fn((id: number) => mockAllEvents.find((event) => event.id === id)),
   findAll: jest.fn(() => mockAllEvents),
   update: jest.fn((mockEventData: UpdateEventInput) => ({ ...mockEventData })),
   delete: jest.fn((id: number) => mockAllEvents.find((event) => event.id === id)),
+  addParticipant: jest.fn((eventId: number, userId: number) => {
```
ys-lin, line 14: Another question, why is mocking the `EventService` necessary? Can't we just call the real `EventService`?

apps/mull-api/src/app/entities/user.entity.ts
```diff
@@ -21,6 +21,10 @@ registerEnumType(RegistrationMethod, {
 @Entity()
 @ObjectType()
 export class User implements IUser {
+  constructor(id?: number) {
```
osahtout, line 24: If I understand correctly, this is needed because the event_participant is a separate table. right?Should id be mandatory?

osahtout (dismissed):
Tested with MySQLWorkBench and graphQL, I can add a user as a participant and see them in the event_particiapants table. No duplicate is found when trying to add the same user to the same event

apps/mull-api/src/app/event/event.service.ts
r-saba: Good question.

`User` is a class in the file `user.entity.ts`. So here `new User(userId)` is instantiating an object from that class. Nothing do with the DB in this line. 

I do this because, later I need to pass a `User` object to the `participants` array which is an array of `User` objects

apps/mull-api/src/app/event/event.service.ts
r-saba: I also had the same question when I was writing this code. The ORM handles it for us, it won't add duplicate users as participants in the database.

apps/mull-api/src/app/event/event.service.ts
ys-lin: Thank you! I have another question, what happens if the user doesn't exist in the database? Will that still get saved into the array?

apps/mull-api/src/app/event/event.service.ts
ys-lin: Oh alright, thank you!

apps/mull-api/src/app/entities/user.entity.ts
r-saba: It's not necessary. I added it because it made our code cleaner. without it, to instantiate a `User` with an id it would take 2 lines like below.
 
The reason I made it optional is because there will be cases where we want to instantiate a new user without an id because the id will be determined by the database. 

```
const user = new User();
user.id = userId
```
vs what I can do now
```
const user = new User(userId)
```

apps/mull-api/src/app/event/event.resolver.spec.ts
r-saba: This file contains unit tests, if we called the real eventService it would save the data to the db. Here I'm just testing that my resolver takes in the expected parameters and returns the expected result.

We're not testing database interactions in our codebase currently.

apps/mull-api/src/app/event/event.service.ts
r-saba: It would throw an error which would have to be handled by the caller of the mutation - on the front-end.

r-saba: By the time this function gets called, the user should already have registered, logged in and be authenticated, so they should already exist in the db. 

If somehow they manage to get past all that and provide an invalid userId, this function would throw an error which would have to be handled by the caller - the front-end.

ys-lin: I'm not very familiar with graphQL so I might be totally wrong, but it's possible to query graphQL without having to go through the front-end right? I think for sanity check, it could be better if we find the user from the DB first( again I'm not very familiar with type ORM, maybe it's not possible to do that) the pseudo code would look like: 
```
user = User.find(id : id) 
if user:
  const event = await this.eventRepository.findOne(eventId, { relations: ['participants'] });
  event.participants.push(user);
 ```
What do you think?

File: apps/mull-api/src/app/entities/user.entity.ts (hunk 5 above)
Comment (osahtout) [reply apps/mull-api/src/app/entities/user.entity.ts:24]: 👍

File: apps/mull-api/src/app/event/event.service.ts (hunk 2 above)
Comment (r-saba) [reply apps/mull-api/src/app/event/event.service.ts:31]: Yeah I know what you mean, I thought about that too. 
My concern was 3 things

1. That will make an additional db call, decreasing performance.
1. Now this function is doing multiple things (finding a user and then adding them to the event).
1. What should we do if a user is not found? Throw an error anyway right, so I don't see the point in doing an extra call to throw an error in the end when TypeORM will throw an error for us anyway.

Comment (ys-lin) [reply apps/mull-api/src/app/event/event.service.ts:31]: Ah, so type ORM actually checks if the user exists, thank you. I might be wrong again, and it really depends on if security is a concern for us, but if anyone can query the graphql api, then anyone can just post an event id and adding whatever user to the participant list...

File: apps/mull-api/src/app/event/event.resolver.spec.ts (hunk 4 above)
Comment (ys-lin) [reply apps/mull-api/src/app/event/event.resolver.spec.ts:14]: Ah thank you! Sorry for having so many questions xD I'm not very familiar with this code base. Not really related to this, just a question. With the codebases I worked with, we have a local test database where we actually write to and clear every time. I'm just wondering if that can be considered in the future?

**Review by ys-lin — DISMISSED** [review]
I just have some questions, if they are not actual concerns, then LGTM 👍 
- [x] Must match Mockup

File: apps/mull-api/src/app/event/event.resolver.ts [hunk 6]
```diff
@@ -31,4 +31,9 @@ export class EventResolver {
   async deleteEvent(@Args('id', { type: /* istanbul ignore next */ () => Int }) id: number) {
     return this.eventService.delete(id);
   }
+
+  @Mutation(() => Event)
+  async addParticipantToEvent(@Args('eventId') eventId: number, @Args('userId') userId: number) {
```
bit172, line 36: Would it be possible to see if the user is already part of the event here? In case there's some edge case we miss. You can reach out to me if you're not sure how to implement it.

apps/mull-api/src/app/event/event.resolver.ts
```diff
@@ -31,4 +31,9 @@ export class EventResolver {
...
+    return this.eventService.addParticipant(eventId, userId);
```
bit172: I've been getting more familiar with GraphQL and I think it would be better to just return a boolean so that we don't have to add fields in the mutation.

apps/mull-api/src/app/event/event.service.ts
bit172: It won't let you add a user that does not exist since it is a foreign key and the user id needs to exist before adding them to an event.

bit172 (changes requested):
LGTM! Joining events works! I just have small things that I would like to change.

apps/mull-api/src/app/event/event.resolver.ts
r-saba: The purpose of this is to add users to events not check if they're in them. If we try to add a user twice it will just add them once and if the user is in the event already it won't do anything. 

Given the above info, is there a benefit to checking if a user is in the event? We could always get the participants list from the event to check if a user is in it or not

apps/mull-api/src/app/event/event.resolver.spec.ts
r-saba: Definitely something we can consider in the future. Maybe when we set up the database for e2e tests we can also set one up for this if we feel it'd be useful

apps/mull-api/src/app/event/event.resolver.ts
bit172: The only benefit I see is blocking the interaction with the database to add the user and maybe having a message to display on the frontend. But I don't think it's necessary (since this interaction would not be possible if the client is implemented properly) and you can keep it as is. 👍

apps/mull-api/src/app/event/event.service.ts
r-saba: @ys-lin Definitely something to discuss in the future.
Maybe we can authenticate the user before running any mutations. However we would need the implementation of JWT in #99 for this. 

That's a bigger issue than this ticket though. We can discuss how to handle that with our team. 

If you'd like to look into the best strategy for authorization and bring it up in one of our meetings go for it, I found this [article](https://www.apollographql.com/docs/apollo-server/security/authentication/) which talks about a few strategies

bit172 (approved):
LGTM!
- [x] Must match Mockup

cindyslittleplanet (approved):
After reading the clarifications comments for the questions, and playing around with it on my side, LGTM! 

- [x] Must match Mockup

osahtout (approved):

cristian-aldea (approved):

Tested it and everything looks good! Thank ragith for the good work

PR #117, opened 2020-12-12
TASK-20: Show Events as a List Back End 

### LINKED ISSUE #75: TASK-20: Show Events as a list Back-end
**Task Description**

As a developer I want to have back-end GraphQL queries to:
- Discover relevant events
- See upcoming events a user has joined
- See events hosted by a user 

**Acceptance Criteria**
- [ ] Events are ordered by event start date
- [ ] Able to query relevant events, upcoming events and events hosted by a user
- [ ] Documentation present for every new component and function where necessary
- [ ] Unit tests implemented for every function

*Hardcoded users for this task is sufficient
*Relevant events can be querying all events for this task

osahtout:
This pull request closes #75

Querying available right now: 
1. All the events where the user is a host
1. All the events where the that the user is co hosting
1. All the  events that the user has joined
1. All the events that the user is not a host, not a co host and has not joined

Note:
For the discover events, the list is returned in no particular order. It is also only returning the events that are public and will not return any event that the restriction is marked as only friends.


apps/mull-api/src/app/event/event.resolver.ts
```diff
@@ -17,6 +17,28 @@ export class EventResolver {
...
+  async participatingEvents(
+    @Args('id', { type: /* istanbul ignore next */ () => Int }) id: number
+  ) {
+    return this.eventService.findJoinedEvents(id);
+  }
+
+  @Query(/* istanbul ignore next */ () => [Event])
+  async discoverEvents(@Args('id', { type: /* istanbul ignore next */ () => Int }) id: number) {
+    return this.eventService.findDiscoverEvent(id);
+  }
```
cristian-aldea: Minor problem but I feel that the wording of the method and args are a bit vague as to what these functions do. Here's my suggestion, you can take parts of it or ignore it, but try to make these easier to understand since it's what is going to be called in the frontend.:
```suggestion
  async hostedEvents(@Args('userId', { type: /* istanbul ignore next */ () => Int }) id: number) {
    return this.eventService.findHostEvents(id);
  }

  @Query(/* istanbul ignore next */ () => [Event])
  async coHostedEvents(@Args('userId', { type: /* istanbul ignore next */ () => Int }) id: number) {
    return this.eventService.findCoHostEvents(id);
  }

  @Query(/* istanbul ignore next */ () => [Event])
  async participatingEvents(
    @Args('userId', { type: /* istanbul ignore next */ () => Int }) id: number
  ) {
    return this.eventService.findJoinedEvents(id);
  }

  @Query(/* istanbul ignore next */ () => [Event])
  async discoverEvents(@Args('userId', { type: /* istanbul ignore next */ () => Int }) id: number) {
    return this.eventService.findDiscoverEvent(id);
  }
```

apps/mull-api/src/app/event/event.service.ts
```diff
@@ -18,6 +20,58 @@ export class EventService {
...
+    });
+  }
+
+  findCoHostEvents(userId: number): Promise<Event[]> {
+    return this.eventRepository
+      .createQueryBuilder('event')
+      .leftJoin('event.coHosts', 'user')
+      .where('user.id = :userId', { userId })
+      .getMany();
+  }
```
cristian-aldea, line 40: Similar to my comment in the resolver
```suggestion
  findHostedEvents(userId: number): Promise<Event[]> {
    return this.eventRepository.find({
      relations: ['host'],
      where: {
        host: {
          id: userId,
        },
      },
    });
  }

  findCoHostedEvents(userId: number): Promise<Event[]> {
    return this.eventRepository
      .createQueryBuilder('event')
      .leftJoin('event.coHosts', 'user')
      .where('user.id = :userId', { userId })
      .getMany();
  }
```

apps/mull-api/src/app/event/event.service.ts
```diff
@@ -18,6 +20,58 @@ export class EventService {
...
+
+  findJoinedEvents(userId: number): Promise<Event[]> {
+    return this.eventRepository
+      .createQueryBuilder('event')
+      .leftJoin('event.participants', 'user')
+      .where('user.id = :userId', { userId })
+      .getMany();
+  }
+
+  findDiscoverEvent(userId: number): Promise<Event[]> {
```
cristian-aldea, line 54: This name doesn't make it easy for me to know what this method does. Could you add some documentation for this method, please?

cristian-aldea (dismissed):
Great Work Omar! I've tested all the functions with a test database and everything works great. Just had a few non-blocking points I'd like addressed if you have the time, mostly related to documentation and clarity

- [x] Events are ordered by event start date
- [x] Able to query relevant events, upcoming events and events hosted by a user
- [x] Unit tests implemented for every function

apps/mull-api/src/app/event/event.service.ts
osahtout: Of course I can. It finds all the events the user has joined

apps/mull-api/src/app/event/event.resolver.ts
osahtout: before I apply these changes, these methods are meant to return all the events of a co host or host. If the new naming implies that I will push the new changes

cristian-aldea: Maybe the "ed" at the end wasn't right, I was trying to find a short wording to describe the behaviour of the method. On second thought the original names you had are okay because I can't think of something that's clearer but still short.

apps/mull-api/src/app/event/event.service.ts
cristian-aldea: Doesn't it find the events that the user has not joined, and is not the host/cohost of?

apps/mull-api/src/app/event/event.resolver.ts
osahtout: Ragith and I couldn't think of any either and we were hoping someone would

apps/mull-api/src/app/event/event.service.ts
osahtout: Oh sorry, I thought I saw findJoinedEvent. Yes you're right

apps/mull-api/src/app/event/event.resolver.ts
cristian-aldea: Yeah disregard this comment

cristian-aldea (approved):

bit172 (approved):
LGTM! It seems that relationships only go one way. I looked into it and we would have to create an entity for each inverse relationship and I wouldn't want that so joining tables is the better option.

- [x] Events are ordered by event start date
- [x] Able to query relevant events, upcoming events and events hosted by a user
- [x] Unit tests implemented for every function

PR #119, opened 2020-12-13
BUG-8: Fix Flaky Test

### LINKED ISSUE #107: BUG-8: Failing and flaky test
**Describe The Bug**

Today, november 30th, the tests in `create-event.spec.tsx` are failing with this error:
image

This is due to the fact that today is the last day of the month of November, and therefore the `span[class="nice-dates-day"]` selector cannot be found. See screenshot
image

This kind of test should either be fixed or even better removed, since its very flaky

**To Reproduce**

1. Run front-end tests during the last day of a month (you can change your computer clock to try it out)

**Expected Behavior**

Tests don't fail when running them on certain days

**Bug Strategy**
- [ ] Fix it now
- [x] Fix it later

cristian-aldea:
closes #107

Instead of querying span[class="nice-dates-day"], which might not exist, the test now queries the first span[class="nice-dates-day -outside"], which corresponds to the first day which is outside the current month, which will always be on the calendar.

To test that the bug doesn't occur anymore, remember that you can change your os clock, which will affect the test

### DIFF
--- a/apps/mull-ui/src/app/pages/create-event/create-event.spec.tsx
+++ b/apps/mull-ui/src/app/pages/create-event/create-event.spec.tsx
@@ -99,7 +99,7 @@ describe('CreateEvent', () => {
 
     let calendarDate = utils.container.querySelector('span[class="nice-dates-day -today"]');
     fireEvent.click(calendarDate);
-    calendarDate = utils.container.querySelector('span[class="nice-dates-day"]');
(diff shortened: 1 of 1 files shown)


JimLau49 (approved):
Thanks for fixing this :)

bit172 (approved):
LGTM!

PR #121, opened 2020-12-16
TASK-28: Set Event Location

### LINKED ISSUE #29: TASK-28: Set Event Location
**User Story Description**

As a host, I want to add the event location to my event so that participants can know where the event is hosted.

**User Acceptance Flow**

1. While creating an event, user has to fill in location field
1. Location field is autocompletes
1. User can choose between two options:
    1. Use current location
    1. Enter an address

**Acceptance Criteria**

- [ ] User acceptance flow is respected
- [ ] Documentation present where needed
- [ ] There must exist an automated Cypress acceptance test in GIF form
- [ ] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=2000%3A222930)

JimLau49:
This PR closes #29

You can try to enter a location by going to  `http://localhost:4200/create-event` and clicking on the location textbox which will open the modal where you can type to get the autocomplete suggestions.

While testing this, make sure your `.env` file is up to date, you can find it on [slack](https://bug-bytes.slack.com/archives/CS5P4NL5Q/p1606083803001700). Both Mull-UI and Mull-API are required to test this PR.

Notes: 

- Added  esModuleInterop to our tsconfig, this allows us to import our axiosInstance (commonJS) module as an es6 module.
- An addition config was added (axiosConfig) which exports axios object into jest, then jest creates a mocked http client so that we can intercept it with our mockedData.
- @bit172  will mention in the comments about how we're using debounce and what it does.


apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.tsx
```diff
@@ -0,0 +1,112 @@
...
+  handleClose: (value: string) => void;
+  input: string;
+}
+
+export default function LocationAutocompleteTextbox({ handleClose, input }) {
+  const [open, setOpen] = useState(false);
+  const [options, setOptions] = useState<string[]>([]);
+  const [getAutocompletedLocations, { loading, data }] = useLazyQuery(AUTOCOMPLETED_LOCATIONS);
+
+  const debounceGetLocation = useMemo(
```
bit172, line 22: I made this function. Here's an explanation:

- debounce
A debounced function will only fire once it has not been called after a set delay. In this example, this function will be called on every keystroke in the autocomplete textbox, but the function will not actually run. Once the user stops typing, it will call the function with the text that was inputted.

- useMemo
Since this is a React component, it will re-render on any change to the state. Therefore, the debounced function will be recreated on every re-render. Using the useMemo hook, we are able to memoize the function so that it is not reset every time.

This essentially cuts down the number of requests being sent as we are only sending a request to the geolocation API once the user has finished typing instead of doing it on every key stroke.

apps/mull-ui-e2e/debug.log
```diff
@@ -0,0 +1 @@
+[1206/123854.351:ERROR:directory_reader_win.cc(43)] FindFirstFile: The system cannot find the path specified. (0x3)
```
cristian-aldea: Is there a reason to have this debug log?

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+    const textbox = utils.getByTestId('location-autocomplete-textbox');
+    fireEvent.change(textbox, { target: { defaultValue: '8485 rue Sherbrooke' } });
+    userEvent.type(
+      screen.getByTestId('location-autocomplete-textbox'),
+      '845 rue Sherbrooke{enter}'
+    );
+
+    const autocompleteTextbox = utils.container.querySelector('#location-input-field') as any;
+    expect(autocompleteTextbox.value).toBe('8485 rue Sherbrooke');
```
cristian-aldea: This test fails when I run it.

```suggestion
    expect(autocompleteTextbox.value).toBe('845 rue Sherbrooke');
```

apps/mull-ui/src/app/components/custom-text-input/custom-text-input.tsx
```diff
@@ -47,6 +53,8 @@ export const CustomTextInput = ({
         type={password ? 'password' : 'text'}
         value={value}
         onChange={onChange}
+        onClick={onClick}
+        readOnly={readOnly}
```
cristian-aldea, line 56: Very nice :)

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import Dialog from '@material-ui/core/Dialog';
+import Search from './location-autocomplete-textbox';
```
cristian-aldea: What you've done here is that you've imported `LocationAutocompleteTextbox` as `Search`, which really confused me initially. I would recommend renaming the component to `LocationSearchBox` and importing it under the same name, to avoid confusion.

If you do rename the component, make sure to also rename the related files (`.tsx`, `.spec.tsx`, and `.scss`)

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
```diff
@@ -0,0 +1,70 @@
...
+    startTime: string;
+    endTime: string;
+    eventTitle: string;
+    description: string;
+    location: string;
+    imageFile: string;
+  }>;
+}
+
+export default function LocationAutocompleteModal({ formik: { touched, setFieldValue, errors } }) {
```
cristian-aldea: You could rename this to `LocationSearchModal`, `LocationAutocompleteModal` is a bit long

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
```diff
@@ -0,0 +1,70 @@
...
+            marginRight: 'auto',
+            marginLeft: '0.2rem',
+            marginTop: '0.8rem',
+            marginBottom: '2rem',
+          }}
+          children={'Edit'}
+          onClick={handleClose}
+          className={'edit'}
+        />
+        <Search handleClose={handleClose} input={inputValue} />
```
cristian-aldea: The way you're using handleClose makes it so that it has to handle two cases. 

The first case is when MullBackButton invokes handleClose. In this case, the passed argument is an event, which is why you need to check for the type of the value.

The second is when Search calls handleClose. In this case, since the value is a string, you also set the input and field values.

You don't need to check for the type and introduce this complexity, you can just remove the if, call handleClose with Search, and give `setOpen(false)` as the `onClick` callback for the back button.

(sorry the suggestion is a bit long, I just did what I described above)

```suggestion
const handleClose = (value: string) => {
    setOpen(false);
    setInputValue(value);
    setFieldValue('location', value);
  };

  return (
    <React.Fragment>
      <CustomTextInput
        title="Location"
        fieldName="location"
        value={inputValue}
        readOnly
        onClick={handleClickOpen}
        hasErrors={touched.location && !!errors.location}
        errorMessage={errors.location}
        svgIcon={<FontAwesomeIcon icon={faMapMarkerAlt} />}
      />

      <Dialog fullScreen open={open}>
        <MullBackButton
          style={{
            marginRight: 'auto',
            marginLeft: '0.2rem',
            marginTop: '0.8rem',
            marginBottom: '2rem',
          }}
          children={'Edit'}
          onClick={() => setOpen(false)}
          className={'edit'}
        />
        <Search handleClose={handleClose} input={inputValue} />
```

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const handleClose = (value) => {
+    setOpen(false);
+    if (typeof value == 'string') {
+      setInputValue(value);
+      setFieldValue('location', value);
+    }
+  };
+
+  return (
+    <React.Fragment>
```
cristian-aldea: You can simply write `<>` and `</>` instead of `React.Fragment`

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.scss
```diff
@@ -0,0 +1,3 @@
+.MuiAutocomplete-endAdornment {
+  width: 0%;
```
cristian-aldea:
```suggestion
  display: none;
```

tsconfig.base.json
```diff
@@ -8,6 +8,7 @@
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "importHelpers": true,
+    "esModuleInterop": true,
```
cristian-aldea, line 11: Which part of the code required this change? If it's in the tests I didn't look into those in much detail so I might have missed it.

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  const handleClickOpen = () => {
+    setOpen(true);
+  };
+
+  const handleClose = (value) => {
+    setOpen(false);
+    if (typeof value == 'string') {
+      setInputValue(value);
+      setFieldValue('location', value);
```
cristian-aldea: Instead of having `latitude: xxx, longitude: xxx` as the field value, we could have something like `Current Location`. I suggest that instead of taking in one value, your `handleClose` function could take in two values, `inputValue`, and `fieldValue`, which will give the Search component more control on what we want to show the user.

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.tsx
```diff
@@ -0,0 +1,112 @@
...
+import TextField from '@material-ui/core/TextField';
+import Autocomplete from '@material-ui/lab/Autocomplete';
+import CircularProgress from '@material-ui/core/CircularProgress';
+import { gql, useLazyQuery } from '@apollo/client';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import './location-autocomplete-textbox.scss';
+import { faMapMarkerAlt, faLocationArrow } from '@fortawesome/free-solid-svg-icons';
+import { debounce } from 'lodash';
+
+export const AUTOCOMPLETED_LOCATIONS = gql`
```
cristian-aldea: One problem I've been having trying out your code is that a few places I'm familiar with (my home, old home, Somerled Park, etc.) don't show up in the search results, which is kind of a bummer. Do you know what might be causing this? Maybe switching API might help?

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.tsx
```diff
@@ -0,0 +1,112 @@
+import React, { useState, ChangeEvent, useEffect, useMemo } from 'react';
```
cristian-aldea: One problem I've been having is that the modal closes whenever I delete all of the content of the text field after an address is already confirmed. This gets really annoying when I'm trying to change an address. I can't delete the content in the text field and write a new address before the modal closes on me.

cristian-aldea (changes requested):
Great work everyone who worked on this, I'm glad I get to see it in action. 

I had a few blocking and non-blocking comments, nothing critical. Please let me know if you need clarifications.

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.tsx
cristian-aldea: I tried the google places api demo and it seems to find the places I mentioned above just fine, might be worth switching too if other people also can't find places they want to search for

https://developers.google.com/maps/documentation/javascript/examples/places-searchbox

bit172: I think this is just the API that wants you to put in the addresses exactly how they are. If the street name is in french ie Rue vs Street, it won't find it.

apps/mull-ui-e2e/debug.log
JimLau49: Gonna remove this

r-saba:
When selecting current location instead of showing `Lat` and `Long` in the input field, it should show "Current Location"

r-saba:
If I submit then go back to the form the location is missing in the input field.

If I do the above I can submit the form with an empty location, the validation didn't stop me.

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.tsx
JimLau49: Yeah i think the intention was that you would always be forced to put in a new valid address but a problem occurs lets say if u wanted to change the "750 street" to "850 street" , the moment u delete "7" it will close the modal making it impossible to ever write 8 in the modal unless u refresh the page. 

This was something i had brought up, and a potential solution was to have a clear (x) button on the autocomplete-textbox (mockups would need to be updated for this small change). Let me know what you guys think.

r-saba:
I thought this was the last ticket for the create-event flow. Shouldn't it add the location to the review page as well and save it in the db?

r-saba:
The API doesn't seem to work too well, I put a park near my house and it couldn't find it. Could we look into a better API?

Very quick googling and I found this API which found the park near my house https://community.algolia.com/places/
Unfortunately they're shutting down in 2022, if we can't use this maybe there's others?

r-saba (changes requested):
looking good, did some functional test and left some comments also, I thought this was the last ticket for the create-event flow. Shouldn't it add the location to the review page as well and save it in the db?

I'll be reviewing the code tomorrow after my final

apps/mull-ui/src/app/pages/create-event/create-event.tsx
```diff
@@ -239,15 +240,7 @@ const CreateEventPage = ({ history }: CreateEventProps) => {
...
-          <CustomTextInput
-            title="Location"
-            fieldName="location"
-            value={formik.values.location}
-            onChange={formik.handleChange}
-            hasErrors={formik.touched.location && !!formik.errors.location}
-            errorMessage={formik.errors.location}
-            svgIcon={<FontAwesomeIcon icon={faMapMarkerAlt} />}
-          />
+          <LocationInput formik={formik} />
```
r-saba: If I submit then use the back button to go back to the form the location is missing in the input field.

If I do the above I can submit the form with an empty location, the validation didn't stop me.

apps/mull-api/axiosConfig.ts
```diff
@@ -0,0 +1,6 @@
+import axios from 'axios';
+const axiosInstance = axios.create({
+  baseURL: `https://app.geocodeapi.io/api/v1/autocomplete?apikey=${process.env.GEOCODE_KEY}&text=test_location&size=5`,
```
r-saba, line 3: The API doesn't seem to work too well, I put a park near my house and it couldn't find it. Could we look into a better API? Or is it another issue?

Very quick googling and I found this API which found the park near my house https://community.algolia.com/places/
Unfortunately they're shutting down in 2022, if we can't use this maybe there's others?

apps/mull-api/src/app/event/location-autocomplete/location-autocomplete.resolver.ts
```diff
@@ -0,0 +1,14 @@
+import { Query, Args } from '@nestjs/graphql';
+import axiosInstance from '../../../../axiosConfig';
+
+export class LocationAutocompleteResolver {
+  @Query(/* istanbul ignore next */ () => [String])
+  async getAutocompletedLocations(@Args('userInput') userInput: String) {
+    const response = await axiosInstance.request({
+      method: 'get',
+      url: `https://app.geocodeapi.io/api/v1/autocomplete?apikey=${process.env.GEOCODE_KEY}&text=${userInput}&size=5`,
```
r-saba: The API doesn't seem to work too well, I put a park near my house and it couldn't find it. Could we look into a better API? Or is it another issue?

Very quick googling and I found this API which found the park near my house https://community.algolia.com/places/
Unfortunately they're shutting down in 2022, if we can't use this maybe there's others?

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
r-saba: Agreed, it should show "Current Location"

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.tsx
cristian-aldea: That's possible, but I legit could find some addresses or places no matter how I wrote them

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.tsx
cristian-aldea: No I don't agree with your solution, the solution would be to not close the modal when the last character is deleted. I don't know how that would be done since I didn't pay too much attention on the implementation of the text box but its 100% possible

apps/mull-api/src/app/event/location-autocomplete/location-autocomplete.resolver.ts
cristian-aldea: Had the same issue as you and left a comment, have a look: https://github.com/RGPosadas/Mull/pull/121#discussion_r544754305

cristian-aldea: The google places api pretty much gets my searches perfectly everytime, just like when I'm searching on google maps. If we were to switch API I think google would be a good option

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.tsx
cristian-aldea: Not saying that having a clear button is a bad idea, but its not needed, and its just a band-aid for a bug that should be easy to fix

apps/mull-ui/src/app/components/custom-text-input/custom-text-input.tsx
ys-lin: thanks :)

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
ys-lin: Sure, we can rename it

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
ys-lin: Is it going to give a type error if the value passed into the function is not a string?

tsconfig.base.json
ys-lin: Please read the description, Jimmy elaborated on that :)

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
ys-lin: That's just a place holder, because we might need long and lat in the future for the backend, and this clearly shows the next person how to get it

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.tsx
ys-lin: Yeah it's just the API

apps/mull-ui/src/app/pages/create-event/create-event.tsx
ys-lin: Could you send a gif of the situation? I'm not exactly sure what you mean

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.tsx
ys-lin: At this point can we just remove the modal? It complicates things 2000% xD

ys-lin: My solution is to have a simple searchable select, and that will remove material UI altogether

apps/mull-ui/src/app/pages/create-event/create-event.tsx
r-saba: I will show you today

apps/mull-api/src/app/event/location-autocomplete/location-autocomplete.resolver.ts
ys-lin: changing the API might mean a lot of changes, for instance, the key might not have be in the backend because google has a way to restrict which domain can access the api. Also, with the frustrations mentioned above, fixing this ticket might not be worth it. If changing the API is necessary, then the ticket should just be restarted. Also, if we can change the implementation of the location autocomplete to not have a modal, but a searchable select instead, it will make this ticket so much easier. What do you think?

apps/mull-ui/src/app/pages/create-event/create-event.tsx
r-saba: @ys-lin I can only show you after the final but it's simple to replicate when you run the app

1. Fill in the form and submit
2. Press edit from review page
*Notice the location is empty now
3. Submit the form
*Notice the form was submitted with empty location

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.tsx
JimLau49: Ok sure, i think we'll go that approach then @TheGreatMarkus, should be a simple fix

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
cristian-aldea: With my change yes, but that's not an issue since MullBackButton wouldn't use the handleClose function anymore

tsconfig.base.json
cristian-aldea: Ah right my bad

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
cristian-aldea: I think my suggestion wasn't clear. I didn't mean that we should change anything about the lat long value, I just want to change what is shown to the user through what shows up in the textbox once they choose to use their current location

apps/mull-ui/src/app/pages/create-event/create-event.tsx
ys-lin: Ah, I got it now, thank you!

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.tsx
bit172: I don't think the modal is an issue. Tweaking the logic would be better.

apps/mull-api/axiosConfig.ts
bit172: Is there a reason to have a base URL set if it's not being used in other places?

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.tsx
ys-lin: yeah but there's also the fact that the location cannot be sent (probably because of formik), and Idk if it's worth the extra time to patch all of this instead of creating a simple searchable select

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
ys-lin: yeah I got what you mean, but I think displaying lat and long for now is going to be clear for the next person who will use lat and long

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
ys-lin: this would be an issue though, since `handle_close` can be called even for null input value

apps/mull-api/axiosConfig.ts
bit172: Also, if this is used only in testing, it shouldn't be at the root of the project.

cristian-aldea:
As discussed in the daily, here is a summary of the issues to be resolved in this ticket:
- [X] Switch to another API
- [x] Validation isn't working
- [ ] Location isn't sent to review page
- [X] Modal closes when trying to clear the text input for location search after a location was set.

bit172 (commented):
Everything that I saw was already mentioned. Just resolve the issues and it should be good. I know you all worked very hard on this and it looks really good for your first issue. Good work!

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.tsx
JimLau49: Changed to Google api these problems should no longer be happening

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
```diff
@@ -0,0 +1,74 @@
...
+  };
+
+  const handleSetValue = (location: ILocation) => {
+    setInputValue(location.title);
+    setFieldValue('location', location);
+    handleClose();
+  };
+
+  return (
+    <React.Fragment>
```
cristian-aldea: You're still using `<React.Fragment>` in your code (not just here). You can change it to `<>` to keep it consistent with the rest of our code

apps/mull-api/axiosConfig.ts
TheGreatMarkus: This wasn't replied to

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
cristian-aldea: Do you mean the next people who will be working on this implementation? I don't think this is a supporting ticket for anything

apps/mull-api/src/app/event/location-autocomplete/location-autocomplete.resolver.ts
```diff
@@ -0,0 +1,14 @@
+import { Query, Args } from '@nestjs/graphql';
+import axiosInstance from '../../../../axiosConfig';
+
+export class LocationAutocompleteResolver {
+  @Query(/* istanbul ignore next */ () => [String])
+  async getAutocompletedLocations(@Args('userInput') userInput: String) {
```
cristian-aldea: This should throw an error when something goes wrong. For example, I forgot to add in the key in my `.env` file, and it took me a while to understand why

apps/mull-api/src/app/entities/location.entity.ts
```diff
@@ -9,7 +9,7 @@ export class Location implements ILocation {
   id: number;
 
   @Column('point')
-  point: string;
+  title: string;
```
cristian-aldea, line 11: why was this renamed? There should be two fields, one called `title` and one called `coordinate` or `point`, which contains the lat/long position of the location.

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
JimLau49: Theres #124 which is gonna handle some of the backend since we broke this ticket down into front-end and back-end tasks since this ticket a full user story before we started breaking tickets down

cristian-aldea:
I'm so sorry, I wasn't done with my review, but for some reason it sent the feedback I was writing. I'll have the rest of my comments soon

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.tsx
cristian-aldea: This was fixed for me just fyi

libs/types/src/lib/types.ts
```diff
@@ -42,9 +42,10 @@ export interface IEvent {
 }
 
 export interface ILocation {
-  id: number;
-  point: string;
-  events?: IEvent[];
+  id?: number;
+  title: string;
+  placeId?: string;
```
cristian-aldea, line 52: Wait what is place id?

tsconfig.base.json
```diff
@@ -8,6 +8,8 @@
...
+    "allowSyntheticDefaultImports": true,
```
cristian-aldea, line 12: What does this setting do?

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
```diff
@@ -0,0 +1,74 @@
...
+            marginRight: 'auto',
+            marginLeft: '0.2rem',
+            marginTop: '0.8rem',
+            marginBottom: '2rem',
+          }}
+          children={'Edit'}
+          onClick={handleClose}
+          className={'edit'}
+        />
+        <LocationAutoCompleteTextbox handleSetValue={handleSetValue} input={inputValue} />
```
cristian-aldea, line 59: I think the textbox and the search results could use some padding on the left and right.

cristian-aldea (changes requested):
Looking better! Had some other small comments and suggestions

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
cristian-aldea: True, but this comment is just related to the front-end, and what is shown to the user. In any case I see it was fixed, so I'll be resolving this comment.

libs/types/src/lib/types.ts
JimLau49: To explain it briefly, the google places api (autocomplete) which we're using doesn't actually give back coordinates, instead it gives back a placeId that we can use to get back the location with geocoding.

https://developers.google.com/maps/documentation/javascript/geocoding

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
JimLau49: Small error on our side this was done but not committed will be done in the next pushes

tsconfig.base.json
JimLau49: So, @bit172 had something like `import * as cookieParser from 'cookieParser'` because it isn't an export default , so we can import it as `import cookieParser from 'cookieParser'` with this. 

With esModuleInterop (which we need to import axios as an es6 import), it would be complaining that cookieParser has no call signatures as isn't default exported and thus imported as commonJS 

image

So this was added to allow us to synthetically import cookieParser as an es6 module

libs/types/src/lib/types.ts
JimLau49: image

apps/mull-api/src/app/entities/location.entity.ts
JimLau49: So this was changed because now title will be the location in common language and coordinates which will contain the coordinates if the user uses current location, otherwise the location's placeId will be used, i left an explanation below as to why we're using placeId instead of [Lat, Long]

libs/types/src/lib/types.ts
JimLau49: So currently the user flow now goes as follows:

User presses Current location --> we'll use geolocation api to get back their [Lat,Long]

User types in an address and selects one --> We get back the string and its associated placeId

cristian-aldea: Looked into this with @JimLau49, we can use https://developers.google.com/places/web-service/search to search for places too, which will give us the lat/lng right off the bat. The search is smart (you don't need to type out the entire address to get the result) but it doesn't autocomplete (you need to type out the street name for example).

Unfortunately, for places autocomplete, there is no option to get the lat/lng, so we have to compromise on something.

apps/mull-api/src/app/entities/location.entity.ts
cristian-aldea: As discussed, the coordinate field will be added in the entity in #124

package.json
```diff
@@ -117,7 +122,7 @@
     "@typescript-eslint/eslint-plugin": "2.19.2",
     "@typescript-eslint/parser": "2.19.2",
     "babel-jest": "26.2.2",
-    "cypress": "^4.1.0",
+    "cypress": "^6.0.1",
```
r-saba, line 127: Just curious is there a reason this was updated, did it add some new features or fix something?

apps/mull-api/src/app/event/location-autocomplete/location-autocomplete.resolver.ts
cristian-aldea: Thank you for implementing my suggestion @JimLau49

package.json
r-saba: Updating package.json can have unexpected consequences, in the future lets have separate commits for this like I did [here](https://github.com/RGPosadas/Mull/pull/129/commits/85b8c6af65094a901146f5d449474846f8de80ee). 

I remember it happened to us in minicap and if we had separate commits it would have been easier to debug and fix

package.json
```diff
@@ -70,6 +74,7 @@
     "passport-twitter": "^1.0.4",
     "react": "16.13.1",
     "react-dom": "16.13.1",
+    "react-geolocated": "^3.1.0",
```
r-saba: Is this being used anywhere in the app?

apps/mull-api/src/app/event/location-autocomplete/location-autocomplete.resolver.ts
```diff
@@ -0,0 +1,18 @@
...
+    const response = await axiosInstance.request({
+      method: 'get',
+      url: `https://maps.googleapis.com/maps/api/place/autocomplete/json?input=${userInput}&key=sadas`,
+    });
+    const autocompleteLocations = response.data.predictions.map((field) => field.description);
+
+    if (response.data.error_message) {
+      throw new Error(response.data.error_message);
+    }
```
cristian-aldea: Really minor, but it doesn't make sense to process the response data before checking for error, the order of these two statements should be switched
```suggestion
    if (response.data.error_message) {
      throw new Error(response.data.error_message);
    }
    
    const autocompleteLocations = response.data.predictions.map((field) => field.description);
```

apps/mull-api/src/app/entities/location.entity.ts
r-saba: Then the column type here is no longer `point` then

Replying here: since this involves the backend it will be addressed in #124

apps/mull-api/src/app/entities/location.entity.ts
```diff
@@ -9,7 +9,7 @@ export class Location implements ILocation {
   id: number;
 
   @Column('point')
```
r-saba: This will set the type to varchar(255) in the database
```suggestion
  @Column()
```

apps/mull-api/axiosConfig.ts
```diff
@@ -0,0 +1,6 @@
+import axios from 'axios';
+const axiosInstance = axios.create({
+  baseURL: `https://maps.googleapis.com/maps/api/place/autocomplete/json?input=845_Rue_Sherbrooke&key=${process.env.GOOGLE_KEY}`,
```
r-saba, line 7: Noticed you're making an instance here but overwriting the URL everytime you call the instance. You could just use the `axios.get()` method instead and you wouldn't need this `axiosConfig.ts` file anymore.

Link to usage in the [docs](https://github.com/axios/axios#note-commonjs-usage)

apps/mull-ui/src/app/pages/create-event/create-event.tsx
```diff
@@ -113,7 +114,16 @@ const CreateEventPage = ({ history }: CreateEventProps) => {
       description: Yup.string()
         .required('Event Description is required.')
         .max(5000, 'Event Description must be under 5000 characters.'),
-      location: Yup.string().required('Event Location is required.'),
+      location: Yup.mixed()
+        .required('Event Location is required.')
+        .test(
```
r-saba, line 101: Is there a reason you couldn't use `Yup.string()` here?

apps/mull-ui/src/app/pages/create-event/create-event.tsx
r-saba: Thanks for fixing this

apps/mull-ui/src/app/pages/create-event/create-event.tsx
```diff
@@ -16,12 +16,13 @@ import {
 import DateCalendar from './date-calendar/date-calendar';
 import { EventPage } from './../event-page/event-page';
 import { useToast } from '../../hooks/useToast';
+import LocationInput from '../create-event/location-autocomplete/location-autocomplete-modal';
```
r-saba: You should import Components with the same name you used as the default export.
```suggestion
import LocationAutocompleteModal from '../create-event/location-autocomplete/location-autocomplete-modal';
```

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.tsx
```diff
@@ -0,0 +1,113 @@
...
+        350
+      ),
+    []
+  );
+
+  useEffect(() => {
+    if (!loading && data) {
+      setOptions(data.getAutocompletedLocations);
+    }
+  }, [loading, data, setOptions]);
```
r-saba, line 38: I knw how useEffect works but what does putting `setOptions` function as a dependency do here?

apps/mull-ui-e2e/src/integration/US-1.1.spec.ts
```diff
@@ -38,8 +38,33 @@ frameSizes.forEach((frame) => {
     it('should type into the event description input', () => {
       cy.get('#description').type('test description').should('have.value', 'test description');
     });
-    it('should type into the event location input', () => {
-      cy.get('#location').type('test location').should('have.value', 'test location');
+
+    it('should type into the event location modal and return autocompleted address', () => {
+      cy.get('#location').click();
+      cy.get('#location-input-field').should('be.visible');
+      const res = '845 Rue Sherbrooke';
```
r-saba: This is not being used anywhere
```suggestion
      const res = '845 Rue Sherbrooke';
```

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.spec.tsx
```diff
@@ -0,0 +1,33 @@
...
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('should open autocomplete modal', async () => {
+    const utils = render(<LocationAutocompleteModal formik={mockFormik} />);
+    const autocompleteModal = utils.container.querySelector('#location');
+    await waitFor(() => {
+      fireEvent.click(autocompleteModal);
+    });
+  });
```
r-saba: This isn't testing anything, there should be an `expect` to check that the modal was opened

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+    );
+
+    const textbox = utils.getByTestId('location-autocomplete-textbox');
+    fireEvent.change(textbox, { target: { defaultValue: '8485 rue Sherbrooke' } });
+    userEvent.type(
+      screen.getByTestId('location-autocomplete-textbox'),
+      '845 rue Sherbrooke{enter}'
+    );
+
+    const autocompleteTextbox = utils.container.querySelector('#location-input-field') as any;
```
r-saba: Here's the correct type for this. Fix it everywhere else you use `any` in the tests please
```suggestion
    const autocompleteTextbox = utils.container.querySelector<HTMLInputElement>('#location-input-field');
```

apps/mull-api/axiosConfig.ts
r-saba: Let me know if you need more clarity, we can hop on a quick call

apps/mull-ui-e2e/src/integration/US-1.1.spec.ts
```diff
@@ -38,8 +38,33 @@ frameSizes.forEach((frame) => {
...
+      cy.get('#location-input-field').type('845 Rue Sherbrooke');
+      cy.get('#location-input-field-option-0', { timeout: 3500 }).should(
+        'have.text',
+        '845 Rue Sherbrooke, Montreal, QC, Canada'
+      );
```
r-saba: This test fails for me because the first result is 845 Rue Sherbrooke **Ouest**, Montreal, QC, Canada. I'm not sure if this was just not updated from the previous API or it depends on the location where the test is being ran. I suggest the following fix to make it less flaky if it's a location based issue
```suggestion
      cy.get('#location-input-field-option-0', { timeout: 3500 }).should(
        'contain.text',
        '845 Rue Sherbrooke'
      );
```

apps/mull-ui-e2e/src/integration/US-1.1.spec.ts
```diff
@@ -78,8 +103,16 @@ frameSizes.forEach((frame) => {
...
 
-      cy.get('#location').type('test location');
+      cy.get('#location').click();
+      cy.get('#location-input-field').should('be.visible');
+
+      cy.get('#location-input-field').type('845 Rue Sherbrooke');
+      cy.get('#location-input-field-option-0', { timeout: 3500 }).should(
+        'have.text',
+        '845 Rue Sherbrooke, Montreal, QC, Canada'
+      );
```
r-saba: Same issue as above here
```suggestion
      cy.get('#location-input-field-option-0', { timeout: 3500 }).should(
        'contain.text',
        '845 Rue Sherbrooke'
      );
```

apps/mull-ui-e2e/src/integration/US-1.1.spec.ts
```diff
@@ -38,8 +38,33 @@ frameSizes.forEach((frame) => {
...
+
+    it('should click current location on event location modal', () => {
+      cy.get('#location').click();
+      cy.get('#location-input-field').should('be.visible');
+      cy.get('#location-input-field-option-0').click();
+      cy.get('#location')
+        .invoke('val')
+        .should('contain', 'longitude')
+        .should('contain', 'latitude');
+    });
```
r-saba: Test needs to be updated since we don't show `longitude` and `latitude` anymore

r-saba (changes requested):
Really good work! Just some more comments to make this PR perfect, thank you!

Saw an unused import or 2, make sure to run the linter and clean those up plus any other complaints it has that are related to your PR.

apps/mull-api/axiosConfig.ts
r-saba: I have the same question https://github.com/RGPosadas/Mull/pull/121#discussion_r547576278

apps/mull-api/axiosConfig.ts
JimLau49: Yeah, initially this was done with `axios.get()` but the reason why theres an axios instance being created is because of the test that intercepts this call using ```jest.mock('../../../../axiosConfig', ...``` . We weren't able to intercept with the api call without an axios instance which is why the `axiosConfig.ts` was created

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
```diff
@@ -0,0 +1,74 @@
...
+    eventTitle: string;
+    description: string;
+    location: ILocation;
+    imageFile: string;
+  }>;
+}
+
+export default function LocationAutocompleteModal({
+  formik: { touched, setFieldValue, errors, values },
+}) {
```
cristian-aldea: You're not using `LocationAutocompleteModalProps `, and losing out on Typescript goodness

See: https://github.com/formium/formik/issues/2023#issuecomment-596663900

Please create a `CreateEventForm` interface like the one below, put it in the `types` library, and set it as the type for the useFormik statement in `create-event.tsx`

In `LocationAutocompleteModalProps`, change the type of `formik` to `FormikContextType<CreateEventForm>`, as shown in the suggestion

```tsx
const formik = useFormik<CreateEventForm>({ // line 86
```

```suggestion
export interface LocationAutocompleteModalProps {
  formik: FormikContextType<CreateEventForm>;
}

export interface CreateEventForm {
  activeRestriction: EventRestriction;
  startDate: Date;
  endDate: Date;
  startTime: string;
  endTime: string;
  eventTitle: string;
  description: string;
  location: ILocation;
  imageFile: string;
}

export default function LocationAutocompleteModal({
  formik: { touched, setFieldValue, errors, values },
}: LocationAutocompleteModalProps) {
```

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const [inputValue, setInputValue] = React.useState('');
+  const [open, setOpen] = React.useState(false);
+
+  const handleClickOpen = () => {
+    setOpen(true);
+  };
+
+  const handleClose = () => {
+    setOpen(false);
+  };
```
cristian-aldea, line 36: These methods aren't needed, just replace the calls to these methods with calls to `setOpen`

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
```diff
@@ -0,0 +1,74 @@
...
+      />
+
+      <Dialog fullScreen open={open}>
+        <MullBackButton
+          style={{
+            marginRight: 'auto',
+            marginLeft: '0.2rem',
+            marginTop: '0.8rem',
+            marginBottom: '2rem',
+          }}
```
cristian-aldea: Avoid inline styles like this if possible

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
```diff
@@ -0,0 +1,74 @@
...
+    location: ILocation;
+    imageFile: string;
+  }>;
+}
+
+export default function LocationAutocompleteModal({
+  formik: { touched, setFieldValue, errors, values },
+}) {
+  const [inputValue, setInputValue] = React.useState('');
+  const [open, setOpen] = React.useState(false);
```
cristian-aldea: For consistency 

```suggestion
  const [inputValue, setInputValue] = useState('');
  const [open, setOpen] = useState(false);
```

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.tsx
```diff
@@ -0,0 +1,113 @@
...
+  handleClose: (value: string) => void;
+  input: string;
+}
+
+export default function LocationAutocompleteTextbox({ handleSetValue, input }) {
+  const [open, setOpen] = useState(false);
+  const [options, setOptions] = useState<string[]>([]);
+  const [getAutocompletedLocations, { loading, data }] = useLazyQuery(AUTOCOMPLETED_LOCATIONS);
+  const CURRENT_LOCATION = 'Current Location';
+  const debounceGetLocation = useMemo(
```
cristian-aldea, line 22:
```suggestion

  const debounceGetLocation = useMemo(
```

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+export interface LocationAutocompleteTextboxProps {
+  handleClose: (value: string) => void;
+  input: string;
+}
+
+export default function LocationAutocompleteTextbox({ handleSetValue, input }) {
+  const [open, setOpen] = useState(false);
+  const [options, setOptions] = useState<string[]>([]);
+  const [getAutocompletedLocations, { loading, data }] = useLazyQuery(AUTOCOMPLETED_LOCATIONS);
```
cristian-aldea: Is there no type we can give to `data`? If not then one should be made

cristian-aldea (changes requested):
Sorry for the multiple reviews on my part, had a few more suggestions for this pr

apps/mull-ui-e2e/src/integration/US-1.1.spec.ts
JimLau49: Yeah this most likely wasn't updated

apps/mull-ui/src/app/pages/create-event/create-event.tsx
JimLau49: Location has been changed to an object, that takes a title and placeId (won't be optional once #124 is done). So it was changed to `mixed`

r-saba: Should we use `Yup.object()` then?

apps/mull-api/axiosConfig.ts
r-saba: I see, maybe we should only use it in tests for now until we figure out how to mock it. And maybe you could add documentation explaining the purpose of this file for future readers and under what conditions it could be removed

apps/mull-ui-e2e/src/integration/US-1.1.spec.ts
r-saba: Ah I see, my suggestion would make this test API/google update independent so I think we should go with it instead of hardcoding values.

JimLau49: Agreed :)

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.tsx
```diff
@@ -0,0 +1,113 @@
...
+    if (navigator.geolocation) {
+      navigator.geolocation.getCurrentPosition((position) => {
+        let pos = `latitude: ${position.coords.latitude}, longitude: ${position.coords.longitude}`;
+        handleSetValue({ title: 'Current Location', coords: pos });
+      });
+    }
+  };
+
+  return (
+    <Autocomplete
```
osahtout: If it's possible, can we  remove the box shadow and make the background white in the auto complete popups 
Screenshot 2020-12-31 114625

ys-lin: sure, I will look into that :)

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.spec.tsx
JimLau49: Addressed

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.spec.tsx
JimLau49: Addressed

apps/mull-ui-e2e/src/integration/US-1.1.spec.ts
JimLau49: Addressed

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
JimLau49: This one is a bit tricky, i tried to create an .scss file and importing it inot the autocomplete-modal with the styling 

```.mull-back-button .edit {
  margin-right: 'auto';
  margin-left: '0.2rem';
  margin-top: '0.8rem';
  margin-bottom: '2rem';
}
```

but it didn't seem to be able to apply the styling to the mull-back button, not sure if this is because theres a style prop inside the mull back button component

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.tsx
JimLau49: This will get a type when we do the graphql codegen which will make it mandatory after @bit172 's PR gets merged so we will address it then

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.tsx
JimLau49: Ok so @ys-lin, @bit172  and I looked into it,  we think that the boxshadow you're talking about is outside of the application as we can't see the same one in the picture in the following image

Additionally, the options appear to have a white background already 

image

We can discuss this further on a call if needed 🚀

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
JimLau49: I think the material UI example was followed when creating this functions, personally im not too sure if they should be removed or kept

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
JimLau49: Not sure if this was already done by @ys-lin  but this is what it looks like to me now

image

package.json
JimLau49: Will do in future commits

apps/mull-api/axiosConfig.ts
JimLau49: Documentation added

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-modal.tsx
osahtout: I think what he means is that the input text should start where the autocomplete text start if that makes sense

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.spec.tsx
```diff
@@ -0,0 +1,73 @@
...
+} from './location-autocomplete-textbox';
+
+const mockHandleClose: (e) => void = jest.fn();
+
+const mockLocationAutocompleteTextboxProps: () => LocationAutocompleteTextboxProps = () => ({
+  handleSetValue: mockHandleClose,
+  input: '',
+});
+
+describe('location autocomplete textbox', () => {
```
r-saba, line 19: These tests are spamming my console with the warning below, do you have an idea how this could be fixed? If not I could take a look. I'm cleaning up the console in another ticket, would suck to have it cluttered up again
image

apps/mull-api/src/app/entities/location.entity.ts
r-saba: This will be figured out in #124 now

apps/mull-ui/src/app/components/mull-back-button/mull-back-button.scss
```diff
@@ -17,4 +17,10 @@
     vertical-align: middle;
     color: $primary-color;
   }
+  &.edit {
+    margin-right: auto;
+    margin-left: 0.2rem;
+    margin-top: 0.8rem;
+    margin-bottom: 2rem;
+  }
```
cristian-aldea, line 25: Why was this added?

r-saba (approved):
Functionally LGTM!

Thanks for your work, this is a large pr due to it being before our process change which resulted in smaller PR's. There may be areas for code improvement but in general it's looking good and I'm okay with getting this merged so #124 can get completed.

- [x] User acceptance flow is respected
- [x] Documentation present where needed
- [x] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=2000%3A222930)

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.spec.tsx
r-saba: If this can't be fixed/won't be looked into in this PR make a bug report so I can see if there's a fix for this later.

apps/mull-ui/src/app/components/event-page-info/event-page-info.tsx
```diff
@@ -52,11 +52,11 @@ export const EventPageInfo = ({
       </div>
       <div className="info-row">
         <FontAwesomeIcon icon={faMapMarkerAlt} className="event-page-icon color-grey" />
-        {/* TODO: Remove placeholder text once location is implemented */}
         <p className="row-text" data-testid="event-location">
-          18 Mull Ave, Pointe-Claire, Montreal
+          {/* TODO: placeholder till TASK-33 is done uncomment the line below to see on preview page*/}
+          {/* {event.location.title} */}
+          Test Location
```
cristian-aldea, line 58: If I recall correctly, I fixed this already, I believe I fixed this already when I was with @JimLau49 and @ys-lin. It seems to have been reverted :(

cristian-aldea (approved):
Great work, I left a few questions for you, otherwise, it's good to go. Better merge it sooner rather than later.

- [x] User acceptance flow is respected
- [x] Documentation present where needed
- [x] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=2000%3A222930)

apps/mull-ui/src/app/pages/create-event/location-autocomplete/location-autocomplete-textbox.spec.tsx
JimLau49: Creating a bug ticket for it ... #160

PR #125, opened 2020-12-20
Refactor Navigation

osahtout:
This pull request aims to update some UI to more closely match the mockups.

1. Updated the homepage subheader to be always visible during scrolling and seen as part of the top navbar
1. Updated the time box UI in the event card to have the number bolded.
1. Updated the time in the event info header to be consistent and have the day in front of the month

Also, @TheGreatMarkus refactored the navigation bars to make more sense. Before, the 'Navbar" would be at the bottom on mobile and at the top on desktop. The "header" would be at the top on mobile and not exist on desktop. Now, there is a top navbar, and a bottom navbar, and they change their appearance based on what the mockup shows

NEWNEWScreenshot 2020-12-31 104726

NEWNEWScreenshot 2020-12-31 104809

NEWNEWScreenshot 2020-12-31 104835

Screen Shot 2020-12-19 at 11 03 50 PM

As for @JimLau49's discovery, we do need to create another ticket because it affects more than just the sub nav:
NEWNEWScreenshot 2020-12-31 104907


cristian-aldea:
Wow I've never need a branch name so long lmao

apps/mull-ui/src/app/components/event-card/event-card.tsx
```diff
@@ -32,8 +32,10 @@ export const EventCard = ({ event, style = {} }: EventCardProps) => {
         alt="Event"
       />
       <div className="event-card-datetime" data-testid="event-card-datetime">
-        <div>{`${month} ${day}`}</div>
+        <div style={{ fontWeight: 'bolder' }}>{day} </div>
+        <div>{month}</div>
         <div>{time}</div>
+        <div>{merīdiem}</div>
```
cristian-aldea: Must the meridiem be on a separate line? I don't think it looks nice. And if it's not placed on a separate line, then it would simplify the logic you added in `apps/mull-ui/src/utilities.ts`

apps/mull-ui/src/app/components/event-card/event-card.scss
```diff
@@ -24,6 +24,7 @@
   border-radius: 0.5rem;
   padding: 0.2rem;
   text-align: center;
+  width: 2.8rem;
```
cristian-aldea: It would be better to add padding instead of hardcoding the width if you want some spacing imo.

apps/mull-ui/src/utilities.ts
```diff
@@ -5,14 +5,22 @@
...
   const timeString = Intl.DateTimeFormat('en-us', {
     hour: '2-digit',
     minute: '2-digit',
-  }).format(date);
-
-  return { year: date.getFullYear(), month, day: date.getDate(), time: timeString };
+  })
+    .format(date)
+    .split(' ');
+  timeString.push(''); //to take the 24h time into account
```
cristian-aldea: One thing that is kinda dangerous with this is that we can't know what `Intl.DateTimeFormat`, it depends on the format accepted in the region the user lives in. This might have edge cases we aren't aware of

apps/mull-ui/src/app/components/subnavigation-bar/subnavigation-bar.scss
```diff
@@ -1,8 +1,23 @@
...
+
+.subnavigation-text-container {
+  position: relative;
+  margin-top: 0.5rem;
+  margin: 0 auto;
+  align-items: center;
+  width: 100%;
   display: flex;
   justify-content: space-evenly;
 }
```
cristian-aldea: This styling is really hacky, what you've done is that you're hiding the box shadow of the top navigation bar with the -1rem top margin. I think it would be easier if we make it part of the top navbar and make it appear only when on the discover page.

This would be a bit complicated however since the top nav bar is the navbar on desktop and the header on mobile. Hence I think we might as well refactor those too so they make more sense. One should always be at the top, and the other should be at the bottom, only appearing on mobile.

What do you think @r-saba

cristian-aldea (changes requested):
Left a few suggestions to simplify stuff, I wanna have a talk with you about some potential refactor to simplify the way the navbars work, since its a bit wonky right now. The wonkiness isn't related to your changes though, but we might as well refactor that too

cristian-aldea:
As discussed, I'll be contributing my suggestions to this branch, and if Omar isn't okay with then we can revert the commits

apps/mull-ui/src/app/components/subnavigation-bar/subnavigation-bar.scss
r-saba: I agree @TheGreatMarkus don't think it would be complicated to do either

r-saba: I know we discussed this and I was told that the navigation should look like the one in desktop on both desktop and mobile and that the mockups would be updated to match. Did this change @osahtout ? Not that I mind but I would have done what @TheGreatMarkus mentioned then.

cristian-aldea:
My contributions can be found on this branch: `refactor/cris-test` if anyone is interested. 

I also took the liberty of refactoring the oddities we have with our nav bars

apps/mull-ui/src/app/components/subnavigation-bar/subnavigation-bar.scss
osahtout: I think there's a huge misunderstanding from me and I must apologize. I honestly thought that it was meant that the sub nav would be a separate component as code instead of it being separate visually from the top-nav (unless there was another discussion that I can't remember). 

I think having the sub nav flushed with the scrollable page would cause the user to have difficulty distinguishing the nav and the event card. 

image

cristian-aldea: The way I did it in my test branch isn't ideal either for the reason you mentioned, and it wasn't the approach I was initially going for. Let's discuss the way to tackle this problem during the daily

cindyslittleplanet:
Just a comment about a small issue found on PR #129 that can be fixed here.
I find that the Tabs (the words) are too small compared to menu buttons and the event cards for the desktop view.
They are currently at size 16, which is the same size as the title of each field to complete on the Create Event page.
So, it feels weird to have the words of the Tabs to be the same size as field titles. 

image

apps/mull-ui/src/app/components/subnavigation-bar/subnavigation-bar.scss
cristian-aldea: To document, my approach is now to have the sub nav bar separate from the top navbar. However, the box shadow for the top navbar is disabled and instead place on the sub nav bar, giving the illusion that they are together.

cristian-aldea:
> Just a comment about a small issue found on PR #129 that can be fixed here.

@cindyslittleplanet yeah this is a valid concern. Originally the text size for the event cards seemed too small for me to read comfortably, so I decided to bump it up a lil bit. Now the tabs should also be made a bit bigger to compensate.

apps/mull-ui/src/app/components/event-card/event-card.tsx
```diff
@@ -32,8 +32,8 @@ export const EventCard = ({ event, style = {} }: EventCardProps) => {
         alt="Event"
       />
       <div className="event-card-datetime" data-testid="event-card-datetime">
-        <div>{`${month} ${day}`}</div>
-        <div>{time}</div>
+        <div style={{ fontWeight: 'bolder' }}>{`${day} ${month.toUpperCase()}`}</div>
```
JimLau49: inline styling

apps/mull-ui/src/app/components/event-card/event-card.tsx
```diff
@@ -32,8 +32,8 @@ export const EventCard = ({ event, style = {} }: EventCardProps) => {
...
+        <div style={{ fontWeight: 'normal' }}>{time.replace(/\s/g, '')}</div>
```
JimLau49: same as above

JimLau49 (changes requested):
Desktop looking good for me but i have some comments about mobile view. 

On various mobile displays im getting the sub-nav bar clumped together as well as 'my events' is appearing line broken instead of on the same line

image

osahtout:
> Desktop looking good for me but i have some comments about mobile view.
> 
> On various mobile displays im getting the sub-nav bar clumped together as well as 'my events' is appearing line broken instead of on the same line
> 
> image


Tried to fix it the best I can but at some point, it will still get too cramped and enter a new line but that's before the bottom nav bleeds out the screen

Screen Shot 2020-12-30 at 11 56 24 AM

cristian-aldea:
> Tried to fix it the best ...

Yeah the text might need to be made a bit smaller to fit on the screen in that case. Maybe another breakpoint is needed.

apps/mull-ui/src/app/components/event-card/event-card.tsx
osahtout: removed

apps/mull-ui/src/app/components/event-card/event-card.tsx
osahtout: removed

osahtout:
> > Tried to fix it the best ...
> 
> Yeah the text might need to be made a bit smaller to fit on the screen in that case. Maybe another breakpoint is needed.

Then that would mean that the navigation button would also need to change with that breakpoint since they're cut of from the screen

apps/mull-ui/src/app/components/event-card/event-card.tsx
osahtout: meridiem is no longer used

apps/mull-ui/src/app/components/event-card/event-card.scss
osahtout: that has been addressed

apps/mull-ui/src/utilities.ts
osahtout: This has been addressed

cristian-aldea:
> Then that would mean that the navigation button would also need to change with that breakpoint since they're cut of from the screen

I agree with this @JimLau49. This issue already exists on develop with the bottom nav bar. It should be addressed in another pr through a new breakpoint imo. I'll create the ticket tonight

apps/mull-ui/src/app/components/subnavigation-bar/subnavigation-bar.scss
cristian-aldea: To clarify, I've changed back to something similar to what omar did, my initial approach ended up brining too much complexity to the styling and possibly make it easy to break in the future if the page layout needs to change.

JimLau49 (dismissed):
Thanks for fixing the suggested changes :)

RGPosadas (changes requested):
Nice work, the UI looks more refined and matches our mockups/our PO's vision better!
I also really appreciate the navigation refactoring, it is much easier to follow now than before 🚀 

I'm blocking due to US-1.8's failing E2E test (for all lines like [this](https://github.com/RGPosadas/Mull/blob/develop/apps/mull-ui-e2e/src/integration/US-1.8.spec.ts#L27)) since we now show DAY first before MONTH.

Otherwise, everything looks great 😄 
image

cristian-aldea:
> Nice work, the UI looks more refined and matches our mockups/our PO's vision better!
> ...

Good catch. I fixed the e2e test and took the time to update the branch with what was done on develop. Have another look @RGPosadas  @JimLau49

cristian-aldea:
e2e tests are fixed, linting and units works as before, have another look when you can @RGPosadas  @JimLau49

RGPosadas (approved):

JimLau49 (approved):

cristian-aldea (approved):

PR #129, opened 2020-12-22
Us 2.1: Show Events as List Discover pages

### LINKED ISSUE #65: US-2.1: Show Events as a list on Discover Page
**User Story Description**

As a user, I want to see a list of events so that I can see events that I can join.

**Extra Description**

The user should be able to:
- Discover relevant events
- See upcoming Events
- See their own events

**User Acceptance Flow**

1. User logs in
1. User lands on the home page, which is the discover page
1. User can choose to see:
   1. events relevant to them
   1. Upcoming events that they joined
   1. The events they created

**Acceptance Criteria**
- [ ] User Acceptance Flow is respected
- [ ] There must exist an automated Cypress acceptance test in GIF form
- [ ] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/?node-id=533%3A6528)

*Hardcoded users for this task is sufficient
*Relevant events can be querying all events for this task

r-saba:
Closes #65 

To properly test this PR you will need to setup your database such that User with id 1

- Is not involved in at least 1 event
- Is a participant in at least 1 event, meaning they're in the `event_participants` table
- Is a host or coHost of at least 1 event, meaning in the `hostId` column there is a value of `1` and/or they're in the `event_cohosts` table

**Notes**
run `npm i`, I updated Apollo Client to the latest version to fix a bug with testing.

Images and Location on the event-card are hardcoded. Our API doesn't return this information yet.

**Additional Notes/Addressing PR changes**
I ensured the queries are always fetching the most up to date list of events. After creating an event there is a warning about updating state on an umounted component. [This is a known issue with Apollo client, no solution has been merged yet](https://github.com/apollographql/apollo-client/issues/6209). I don't see any performance or application issues as a result of this warning


cindyslittleplanet:
The tabs (the words) are too small compared to the buttons in the nav bar for the desktop view.
They are currently at size 16, which is the same size as the headers of each field to complete on the Create Event page. 

image

apps/mull-ui/src/app/pages/home/home-discover.scss
```diff
@@ -0,0 +1,14 @@
+.react-swipeable-view-container > div[aria-hidden='true'] {
+  max-height: calc(100vh - 11.5rem);
+  overflow-y: hidden !important; // hides scroll bar when changing routes. !important needed to overwrite the element style set by react swipeable views package.
```
leojrsilao: I don't see the y-scrollbar without ```overflow-y: hidden !important``` when changing from the discover route to upcoming route or my events route unless it happens in a certain browser, I'm using google chrome. Also, there's probably a way to avoid using !important and override the react swipeable view package by using a class name or some property

cindyslittleplanet (commented):
The tabs (the words) are too small compared to the buttons in the nav bar for the desktop view.
They are currently at size 16, which is the same size as the headers of each field to complete on the Create Event page.
And the space between the top and bottom of the tabs is not even. 
image

leojrsilao (dismissed):
Good work! LGTM, the data is now queried for discovering events, upcoming events, and my events. 🚀 Just had a minor comment.

cindyslittleplanet (dismissed):
Since the issue I brought up can be fixed in PR #125 and was not part of the requirements of this US, this looks good to me now! 

- [x] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/?node-id=533%3A6528)

RGPosadas (commented):
Everything currently works well, nice job! 👍 
The only issue I saw was that you would still need to do a manual browser refresh in order to get the updates from the db ([example vid](https://drive.google.com/file/d/1OK_M9o7vpIdd1_zl31gtZBTFUz1V3NRN/view?usp=sharing)). Does this happen to anyone else as well?

There's also a linting error on your E2E test.

RGPosadas (changes requested):
Blocking due to what I wrote above.

apps/mull-ui/src/app/app.spec.tsx
```diff
@@ -1,16 +1,18 @@
...
-import App from './app';
+import { MockedProvider } from '@apollo/client/testing';
 
 describe('App', () => {
   it('should render successfully', () => {
     const { baseElement } = render(
-      <BrowserRouter>
-        <App />
-      </BrowserRouter>
+      <MockedProvider>
```
cristian-aldea, line 11: Just a question. what does `MockedProvider` do?

cristian-aldea:
@RGPosadas yes this happens to me too, @r-saba what do you think about this issue?

apps/mull-ui/src/app/components/event-card/event-card.tsx
```diff
@@ -22,7 +22,7 @@ export const EventCard = ({ event, style = {} }: EventCardProps) => {
   // TODO: Implement distance calculation
   const distance = 15;
 
-  const { day, month, time } = formatDate(event.startDate);
+  const { day, month, time } = formatDate(new Date(event.startDate));
```
cristian-aldea, line 22: I ran into this issue too where `startDate` and `endDate` aren't really dates but are rather strings. Maybe the event service should be tweaked to fix this, so that querying for dates doesn't give back strings. What do you think @r-saba ?

apps/mull-ui/src/app/pages/home/discover/discover.tsx
```diff
@@ -0,0 +1,36 @@
...
+      id
+      title
+      description
+      startDate
+      endDate
+    }
+  }
+`;
+
+export const DiscoverPage = () => {
```
cristian-aldea: `discover.(tsx|spec.tsx|scss)` files should be renamed to better fit the `DiscoverPage` name. Something like `discover-page`

apps/mull-ui/src/app/pages/home/my-events/my-events.tsx
```diff
@@ -0,0 +1,44 @@
...
+      description
+      startDate
+      endDate
+    }
+  }
+`;
+
+export const MyEventsPage = () => {
+  const { data } = useQuery<MyEventData>(GET_PARTICIPATING_EVENTS, {
+    variables: { UserId: 1 },
```
cristian-aldea, line 36: Make sure to leave a `//TODO` for this, and all other places where the user id is hardcoded right now

libs/types/src/lib/types.ts
```diff
@@ -31,8 +31,8 @@ export const EventRestrictionMap = ['Everyone', 'Friends', 'Invite Only'];
 export interface IEvent {
   id: number;
   title: string;
-  startDate: Date;
-  endDate: Date;
+  startDate: Date | string;
+  endDate: Date | string;
```
cristian-aldea: Do you really think this is a good idea? I'd rather not have to deal with composites types. It lead to some messy code in minicap.

cristian-aldea (changes requested):
Everything works great!

- [X] User Acceptance Flow is respected
- [X] Documentation present where necessary
- [X] Unit and UI tests are implemented for every component and function
- [X] There must exist an automated Cypress acceptance test in GIF form
- [X] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/?node-id=533%3A6528)

However, I had a few small blocking concerns:
* Please fix the linting issues you introduced, there's one in `event-card.tsx` and one in `US-2.1.spec.ts`
* you made it so that dates can either be `string` or `Date` types. And I just wanted to have a discussion about this before this is merged. I left a comment about that so you can reply to that comment.

apps/mull-ui/src/app/pages/home/home-discover.scss
r-saba: Did you have enough events to scroll through? It did make a difference for me.
As for the !important, I needed it because there's inline style on the component and this can only be overridden by classes when you use `!important`

This is one of the acceptable cases to use `!important`, [here's some documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) on it, checkout out the "How !important can be used:" section

apps/mull-ui/src/app/app.spec.tsx
r-saba: It mocks the `ApolloProvider` used in `main.tsx`

apps/mull-ui/src/app/components/event-card/event-card.tsx
r-saba: The API gives us a serialized object and it's the responsibility of the front-end to deserialize it. This is why I added the `Date | string` type to the interface below https://github.com/RGPosadas/Mull/pull/129#discussion_r548065726. 

Let's continue this conversation below

libs/types/src/lib/types.ts
r-saba: Agreed, this was the one part I wasn't sure about. An option I was considering is another interface such as `ISerializedEvent` to store the returned data and then once we seralize the event we can use `IEvent`. What do you think, any suggestions?

r-saba: I addressed this by adding a separate `ISeralizedEvent` interface to be used for the serialized event. I think this is the best solution since apollo client does not support automatic reverse seralization at the moment as per their [docs](https://www.apollographql.com/docs/apollo-server/schema/scalars-enums/#custom-scalars).

cristian-aldea: I'm fine with this. I realized that my suggestion to update the IEvent interface to have `startDate` and `endDate` as strings might not be possible. Can you confirm that it's not possible, or did you prefer to have a separate interface for this?

r-saba: Yes exactly, the `IEvent` is used in our API so I had to make a separate interface for our front-end

cristian-aldea (dismissed):
Thanks for addressing my comments. Everything looks good to me.

libs/types/src/lib/types.ts
cristian-aldea: Cool, sorry for being so insistent on that solution despite it not really being feasible haha.

RGPosadas (dismissed):
Great job, LGTM! 🚀 

- [X] User Acceptance Flow is respected
- [X] Documentation present where necessary
- [X] Unit and UI tests are implemented for every component and function
- [X] There must exist an automated Cypress acceptance test in GIF form
- [X] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/?node-id=533%3A6528)

RGPosadas (changes requested):
Blocking due to the failing unit tests on the discovery page sub-routes. Otherwise, linting, e2e, and pre-commit are all clean!

Screenshot of one of the tests:
image

RGPosadas (approved):
I forgot to run `npm install` so I didn't have the package that the unit test needed, so my previous blocking comment is irrelevant. 
LGTM!

cindyslittleplanet (approved):
Still looking good to me!

cristian-aldea (approved):

PR #130, opened 2020-12-29
TASK-36: Event Preview Page

### LINKED ISSUE #128: TASK-36: Event Preview Page
**Task Description**

As a developer, I want to use the event-page component and create the Event Preview page so that users are able to see all the details of an event.

**Subtasks**

- [ ] 36.1 Route to events/id when you click on an event card
- [ ] 36.2 Implement a route for events/id which will show you the Event Preview page
- [ ] 36.3 Query an event by id when event page is used in `Preview` mode
- [ ] 36.4 Implement conditional behaviour for `Preview` mode

**Acceptance Criteria**

- [ ] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=533%3A6528)

leojrsilao:
This PR closes #128 

Both Mull-UI and Mull-API are required to test this PR. 
mull-ui: ```npm run start mull-ui``` 
mull-api: ```npm run start mull-api```

***Steps***
Note: you need to have at least 2 events in your database to test this PR:

1. Once the application is running, go to `http://localhost:4200/events/1`
2. Observe that the event information displayed is based on the event with id 1
3. Update the URL with event id 2 (EG: ```http://localhost:4200/events/2```)
4. Observe that the event information displayed is based on the event with id 2

### DIFF
--- a/apps/mull-ui/src/app/app.tsx
+++ b/apps/mull-ui/src/app/app.tsx
@@ -50,9 +50,8 @@ export const App = () => {
           path="/events/:id"
           children={
             <EventPage
-              event={dummyEvent}
               eventImageURL={'https://i.ytimg.com/vi/-Mb3FoIlTCY/maxresdefault.jpg'}
-              prevPage="home"
+              prevPage="Discover"
             />
           }
         />
--- a/apps/mull-ui/src/app/app.tsx
+++ b/apps/mull-ui/src/app/app.tsx
@@ -50,6 +50,7 @@ export const App = () => {
           path="/events/:id"
           children={
             <EventPage
+              // TODO: Remove placeholder once file download is done
               eventImageURL={'https://i.ytimg.com/vi/-Mb3FoIlTCY/maxresdefault.jpg'}
               prevPage="Discover"
             />
--- a/apps/mull-ui/src/app/app.tsx
+++ b/apps/mull-ui/src/app/app.tsx
@@ -15,6 +15,7 @@ import 'react-toastify/dist/ReactToastify.min.css';
 import './app.scss';
 import { dummyEvent } from '../constants';
 import DiscoverPage from './pages/discover/discover';
+import EventPage from './pages/event-page/event-page';
 
 /**
  * Main component of the application
@@ -44,6 +45,17 @@ export const App = () => {
             <EventCard event={dummyEvent} style={{ marginBottom: '1rem' }} />
           </div>
         </Route>
+        <Route
+          exact
+          path="/events/:id"
+          children={
+            <EventPage
+              event={dummyEvent}
+              eventImageURL={'https://i.ytimg.com/vi/-Mb3FoIlTCY/maxresdefault.jpg'}
+              prevPage="home"
+            />
+          }
+        />
         <Route path={ROUTES.HOME}>
           <div className="page-container">
             <SubNavigationBar />
--- a/apps/mull-ui/src/app/components/event-card/event-card.scss
+++ b/apps/mull-ui/src/app/components/event-card/event-card.scss
@@ -14,7 +14,6 @@
 
 .event-card-image {
   object-fit: cover;
-  cursor: pointer;
 }
 
 .event-card-datetime {
@@ -84,7 +83,6 @@
 .event-card-title {
   font-weight: bold;
   margin-bottom: 0.3rem;
-  cursor: pointer;
 }
 
 .event-card-location {
--- a/apps/mull-ui/src/app/components/event-card/event-card.scss
+++ b/apps/mull-ui/src/app/components/event-card/event-card.scss
@@ -14,6 +14,7 @@
 
 .event-card-image {
   object-fit: cover;
+  cursor: pointer;
 }
 
 .event-card-datetime {
@@ -83,6 +84,7 @@
 .event-card-title {
   font-weight: bold;
   margin-bottom: 0.3rem;
+  cursor: pointer;
 }
 
 .event-card-location {
--- a/apps/mull-ui/src/app/components/event-card/event-card.tsx
+++ b/apps/mull-ui/src/app/components/event-card/event-card.tsx
@@ -13,9 +13,10 @@ import EventMembers from '../event-members/event-members';
 export interface EventCardProps {
   event: Partial<IEvent>;
   style?: React.CSSProperties;
+  onClick?: (event: React.MouseEvent<HTMLDivElement, MouseEvent>) => void;
 }
 
-export const EventCard = ({ event, style = {} }: EventCardProps) => {
+export const EventCard = ({ event, style = {}, onClick }: EventCardProps) => {
   // TODO; set joined based on if current user is part of event
   const [joined, setJoined] = useState<boolean>(false);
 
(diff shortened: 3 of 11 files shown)


apps/mull-ui/src/app/components/event-card/event-card.tsx
```diff
@@ -30,6 +31,7 @@ export const EventCard = ({ event, style = {} }: EventCardProps) => {
         // TODO: Replace placeholder
         src="https://www.citywindsor.ca/residents/parksandforestry/City-Parks/PublishingImages/Assumption%20Park%20Street%20View.JPG"
         alt="Event"
+        onClick={onClick}
```
cristian-aldea: I did some experimentation and found a successful way to handle nested onclicks. just add a call to `e.stopPropagation` to the child `onClick`. This will allow you to set `onClick` on the event-card-container div without shadowing the join button.

Source: https://stackoverflow.com/a/44980950

Here's my suggestion:
```tsx
// 1. Remove the onclick you added

// 2. modify the container div as such:
<div className="event-card-container" style={style} onClick={onClick}>
...

// 3. Modify the join button onClick as such:
<button
  onClick={(e) => {
    e.stopPropagation();
    setJoined(!joined);
  }}
  className={`event-card-join ${joined ? 'joined' : ''}`}
>
```

apps/mull-ui/src/app/app.tsx
```diff
@@ -44,6 +45,16 @@ export const App = () => {
             <EventCard event={dummyEvent} style={{ marginBottom: '1rem' }} />
           </div>
         </Route>
+        <Route
+          exact
+          path="/events/:id"
+          children={
+            <EventPage
+              eventImageURL={'https://i.ytimg.com/vi/-Mb3FoIlTCY/maxresdefault.jpg'}
```
cristian-aldea, line 54: Add a TODO since this placeholder will need to be removed once the file download is implemented.
```suggestion
              // TODO: Remove placeholder once file download is done
              eventImageURL={'https://i.ytimg.com/vi/-Mb3FoIlTCY/maxresdefault.jpg'}
```

apps/mull-ui/src/app/components/event-card/event-card.scss
```diff
@@ -14,6 +14,7 @@
 
 .event-card-image {
   object-fit: cover;
+  cursor: pointer;
```
cristian-aldea: Instead of setting `cursor: pointer` on specific elements, it might be better to create a `.button` scss class and add it to `styles.scss`. We already have some custom styling for `button` elements found in that file, so you can apply the same styling to the `.button` class:

```scss
// in styles.scss
button, .button {
  cursor: pointer;
  outline: none;
  background-color: transparent;
}
```

apps/mull-ui/src/app/pages/event-page/event-page.tsx
```diff
@@ -18,8 +22,33 @@ export const EventPage = ({
...
   onButtonClick,
   eventImageURL,
+  isReview = false,
 }: EventPageProps) => {
-  return (
+  let { id }: any = useParams();
+  const eventId = parseInt(id);
+
+  const GET_SPECIFIC_EVENT = gql`
+    query findSpecificEvent($eventId: Int!) {
```
cristian-aldea:
```suggestion
  const GET_EVENT_BY_ID = gql`
    query getEventById($eventId: Int!) {
```

If you're going with this new name, make sure to rename using F2 so you don't need to rename all the uses manually.

apps/mull-ui/src/app/pages/event-page/event-page.tsx
```diff
@@ -18,8 +22,33 @@ export const EventPage = ({
...
+        id
+        title
+        description
+        startDate
+        endDate
+        restriction
+      }
+    }
+  `;
+  const { loading, error, data } = useQuery(GET_SPECIFIC_EVENT, {
```
cristian-aldea: One problem is that if an error is returned by the backend, there should be an error message in the console.

apps/mull-ui/src/app/pages/event-page/event-page.tsx
```diff
@@ -18,8 +22,33 @@ export const EventPage = ({
...
+    variables: { eventId },
```
cristian-aldea, line 43: If the event doesn't exist in the database, a guard should redirect the user to a 404 page or something. This could be done in a separate ticket, however, to keep this pr small.

cristian-aldea (changes requested):
Lookin good Leo, had some suggestions for you however.

**Subtasks**

- [x] 36.1 Route to events/id when you click on an event card
- [x] 36.2 Implement a route for events/id which will show you the Event Preview page
- [x] 36.3 Query an event by id when event page is used in `Preview` mode
- [x] 36.4 Implement conditional behaviour for `Preview` mode

**Acceptance Criteria**

- [x] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=533%3A6528)

apps/mull-ui/src/app/components/event-card/event-card.tsx
```diff
@@ -24,7 +25,7 @@ export const EventCard = ({ event, style = {} }: EventCardProps) => {
 
   const { day, month, time } = formatDate(event.startDate);
   return (
-    <div className="event-card-container" style={style}>
+    <div className="event-card-container" onClick={onClick} style={style}>
```
cristian-aldea: You can add the `button` class here so that the cursor becomes a pointer like you wanted

cristian-aldea (dismissed):
Everything looks good to me. Thanks for addressing my comments leo

🚀

osahtout (changes requested):
I notice that the picture shown in the preview is the same for all events. I created an event with a new image and that same image sill appears:

Screen Shot 2020-12-31 at 2 23 46 PM


Screen Shot 2020-12-31 at 2 24 27 PM

cristian-aldea:
@osahtout  You're right, but fetching images from the backend isn't part of the scope of this ticket I think.

osahtout:
> @osahtout You're right, but fetching images from the backend isn't part of the scope of this ticket I think.

I wasn't (still not sure) if this is a front-end only ticket since we are querying all the rest of the info

cristian-aldea:
> I wasn't (still not sure) if this is a front-end only ticket since we are querying all the rest of the info

The implementation for this ticket is only in the frontend. Also, the AC for #128 doesn't mention setting up image querying.

osahtout:
> The implementation for this ticket is only in the frontend. Also, the AC for #128 doesn't mention setting up image querying.

oh my bad I thought this criterion (36.3 Query an event by id when event page is used in Preview mode) meant querying everything related to the event including location and image

cristian-aldea:
You're right it's ambiguous. I thought it'd be better to tackle image fetching in a separate ticket since it's pretty complicated I think.

osahtout:
> You're right it's ambiguous. I thought it'd be better to tackle image fetching in a separate ticket since it's pretty complicated I think.

I'm all about separating tickets but I honestly didn't know it was intended to be in a separate task

cristian-aldea:
There was no explicit discussion about this. I guess the issue was that none of the tickets we've worked on have tackled this feature we need. Image fetching was always left as a placeholder or something to do later.

osahtout:
Sure but I don't think it's a good idea to assume a feature will be done later when there was no discussion are no tickets about that feature

cristian-aldea:
Yup its good to document this stuff, I create a ticket to address this, which will be further refined during the meeting today:
https://github.com/RGPosadas/Mull/issues/133

cristian-aldea (approved):

osahtout (approved):
since Images queries are now a separate ticket, I can approve

PR #138, opened 2021-01-04
TASK-31: Local Login Frontend Logic

### LINKED ISSUE #113: TASK-31: Local Login Frontend Logic
**Task Description**

As a developer, I want local logic in the front-end so that the backend is connected.

Users can log in using email and password. This can be achieved using a GraphQL mutation since it is our own implementation of authentication.

**Sub-Tasks**
- [ ] 31.1: Users can login with local accounts
- [ ] 31.2: Users are redirected to the homepage after logging in

**Acceptance Criteria**

cristian-aldea:
This PR closes #113 

This PR implements the local login logic for both the register and login pages. Whether by registering a new user, or logging in with an existing user, the accessToken and userId in the UserContext will be set, allowing the user to access the rest of the application.

**How to test**:

**Registering a new user**:
1. Uncomment this line in `app.tsx` at line 53. :
```tsx
{/* {!accessToken ? <Redirect to={ROUTES.LOGIN} /> : null} */}
```
This line breaks e2e tests, but it will allow you to know if you are logged in or not since it will redirect you to the login page when you're not logged in. This redirection on non-authorized routes will be addressed properly in #99 

2. Run `npm i`
3. Run the backend and frontend applications with `nx serve mull-ui` and `nx serve mull-api`
4. go to `http://localhost:4200/register`
5. Fill out the form to create a new user
6. You should be redirected to the home page, and able to browse the application

**Logging in with an existing user**:
1. Go to `http://localhost:4200/login`
2. Fill out the login form with the user you created
3. You should be redirected to the home page, and able to browse the application

**Accessing the application while not logged in**:
1. If you try to access any routes other than the `/register` and `/login` routes, you should be redirected to the register page.


apps/mull-ui/src/app/pages/event-page/event-page.tsx
```diff
@@ -49,7 +49,7 @@ export const EventPage = ({
   }
 
   if (error) {
-    console.log(error);
+    console.error(error);
```
JimLau49, line 52: Nice catch

apps/mull-ui/src/app/pages/register/register.tsx
```diff
@@ -24,11 +27,14 @@ export const CREATE_USER = gql`
 `;
 
 const Register = ({ history }: RegisterProps) => {
+  const { setUserId, setAccessToken } = useContext(UserContext);
+
   // GraphQL mutation hook to create user
```
JimLau49, line 28: This is a good documentation, i was wondering what this was at first

JimLau49 (approved):
Login and Registering working without issues for me, when not logged in im redirected to the correct page. LGTM 🚀 

- [x] Users can login with local accounts
- [x] Users are redirected to the homepage after logging in

apps/mull-ui/src/app/pages/register/register.tsx
```diff
@@ -42,20 +48,32 @@ const Register = ({ history }: RegisterProps) => {
...
+      }
+
+      loginUser(
+        login,
+        { email: values.email, password: values.password },
+        updateToast,
+        setAccessToken,
+        setUserId,
+        history
+      );
```
bit172: Don't think it's a good idea to login after registering. Since the password basically becomes useless if you can just create an account and access the app.

apps/mull-ui/src/utilities.ts
```diff
@@ -16,3 +22,46 @@ export const formatDate = (
...
+ * @see LoginPage
+ */
+export const loginUser = async (
+  login,
+  loginInput: ILoginForm,
+  updateToast: (type: TypeOptions, message: string) => void,
+  setAccessToken: React.Dispatch<React.SetStateAction<string>>,
+  setUserId: React.Dispatch<React.SetStateAction<number>>,
+  history: History
+) => {
```
bit172: I feel like this function has way too many parameters and I assume you pulled it out to account for the fact that you're using it in different places. IMO, I think that login logic should only belong to the login component.

bit172 (changes requested):
Looks good. Just a small thing about logging in after registering.

apps/mull-ui/src/utilities.ts
```diff
@@ -16,3 +22,46 @@ export const formatDate = (
...
+  if (errors) {
+    const messages = errors.map((err) => err.message).reduce((err1, err2) => err1 + err2);
+    updateToast(toast.TYPE.ERROR, `Error during login: ${messages}`);
+    return;
+  }
+
+  const accessToken = data.login.accessToken;
+  setAccessToken(accessToken);
+
+  const decodedToken = jwtDecode(accessToken) as { id: number };
```
bit172: You should check if the `decodedToken` is valid. Since a falsy or malformed token will throw an InvalidTokenError error. I suggest a try-catch block

apps/mull-ui/src/app/pages/register/register.tsx
cristian-aldea: I guess what could be done is that the user is redirected to the login page to login with their new account.

I'm good with that, I will implement that change

apps/mull-ui/src/utilities.ts
cristian-aldea: > I assume you pulled it out to account for the fact that you're using it in different places

You're right on the money. Since I'll no longer login the user from the register page, I'll be removing this method.

> IMO, I think that login logic should only belong to the login component.

I will be moving this logic back to the login page.

apps/mull-ui/src/utilities.ts
cristian-aldea: Oh good point, I completely forgot about that, will do

bit172 (approved):
Good Work!

- [x] Users can login with local accounts
- [x] Users are redirected to the homepage after logging in

PR #139, opened 2021-01-05
TASK-38: Remove participant from events

### LINKED ISSUE #136: TASK-38: Implement GraphQL mutation to remove participant from event
**Task Description**

As a developer, I want to add a GraphQL mutation to remove participants from an event.

**Acceptance Criteria**

leojrsilao:
This PR closes #136 

To properly test this PR you will need to set up your database such that a user with an id that exists in the user table (EG: userId 1)

- Is a participant in at least 1 event, meaning they're in the event_participants table

Steps to test: 
1. Run `npm run start mull-api`
1. Go to http://localhost:3333/graphql
1. Use the removeParticipantToEvent mutation to remove a participant from a specific event. EG: ` mutation {
      removeParticipantToEvent(eventId:1, userId: 8)
}` (Given that an event with id 1 and user with id 8 exists in your database)

### DIFF
--- a/apps/mull-api/src/app/event/event.service.ts
+++ b/apps/mull-api/src/app/event/event.service.ts
@@ -92,6 +92,14 @@ export class EventService {
     return await this.eventRepository.save(event);
   }
 
+  async removeParticipant(eventId: number, userId: number) {
+    const event = await this.eventRepository.findOne(eventId, { relations: ['participants'] });
+    const user = event.participants.find((participant) => participant.id == userId);
+    const index = event.participants.indexOf(user);
+    event.participants.splice(index, 1);
+    return await this.eventRepository.save(event);
+  }
+
   async delete(id: number): Promise<Event> {
     const event = await this.findOne(id);
     await this.eventRepository.delete(event.id);
--- a/apps/mull-api/src/app/event/event.service.ts
+++ b/apps/mull-api/src/app/event/event.service.ts
@@ -94,10 +94,18 @@ export class EventService {
 
   async removeParticipant(eventId: number, userId: number) {
     const event = await this.eventRepository.findOne(eventId, { relations: ['participants'] });
+    if (event == undefined) {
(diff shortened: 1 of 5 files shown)


apps/mull-api/src/schema.gql
```diff
@@ -60,6 +60,7 @@ type Mutation {
   deleteEvent(id: Int!): Event!
   deleteUser(id: Int!): User!
   login(loginInput: LoginInput!): LoginResult!
+  removeParticipantToEvent(eventId: Float!, userId: Float!): Boolean!
```
RGPosadas: I suggest renaming this method and all other instances of it
```suggestion
  removeParticipantFromEvent(eventId: Float!, userId: Float!): Boolean!
```

apps/mull-api/src/app/event/event.service.ts
```diff
@@ -92,6 +92,14 @@ export class EventService {
     return await this.eventRepository.save(event);
   }
 
+  async removeParticipant(eventId: number, userId: number) {
+    const event = await this.eventRepository.findOne(eventId, { relations: ['participants'] });
+    const user = event.participants.find((participant) => participant.id == userId);
```
RGPosadas, line 101: A check needs to be added here in case userId does not exist in the table since it returns `-1` if userId is not found.
Currently, if the participant DOES NOT exist but the event DOES exist, it will delete the last person that was added to the event. Essentially, it will call `event.participants.splice(-1, 1);`
If possible, ensuring that there is a test added for this would be nice!

RGPosadas (changes requested):

RGPosadas (approved):
LGTM 🚀 
- [x] GraphQL mutation to remove participant from event exists

cindyslittleplanet (approved):
Everything is looking good to me! 
- [x] GraphQL mutation to remove participant from event exists

PR #147, opened 2021-01-13
FIX: Refactor backend method names

### LINKED ISSUE #146: FIX: Rename backend resolver and service methods
Currently, our naming scheme for the backend methods in our resolvers and services is inconsistent. Also, the related names of services and resolvers (for example, a service method which is called in a resolver) don't have the same names sometimes, making it harder to keep track of things.

**Acceptance Criteria**

* Methods in resolvers and services have a consistent naming scheme, and related methods in resolvers/services have the same name.

cristian-aldea:
closes #146 

NOTE: This is a relatively small PR, most of the line changes are due to renames and from generated files. No changes in the app behaviour or bugs should occur. If you can run the tests and app without issues then everything should be good.

This PR aims to standardize the naming of the methods and queries in the backend.

Here's a summary of what was done (see commit messages for details):
* some resolver method names were simplified to make queries easier to write
* Some service method names were made longer so that its behaviour is easier to understand
* Queries in the front-end were adjusted to follow these new names
* some heavy logic found in some resolver methods was moved to the corresponding service methods. The tests were adjusted accordingly. as a reminder, the service should contain the business logic of a query, while the resolver should remain relatively light.
* `createEventInput`, `updateEventInput`, `createUserInput`, etc... variables names were renamed to be simply `user` or `event`. I did this to reduce redundancy since, for example, when you're writing a createEvent query you should know that your input is for the create event query, there's no need to reiterate. Also, the input object you're passing is technically an event, or at least partially one.

### DIFF
--- a/apps/mull-api/src/app/event/event.service.ts
+++ b/apps/mull-api/src/app/event/event.service.ts
@@ -12,34 +12,34 @@ export class EventService {
     private eventRepository: Repository<Event>
   ) {}
 
-  findAll(): Promise<Event[]> {
+  events(): Promise<Event[]> {
     return this.eventRepository.find();
   }
 
-  findOne(id: number): Promise<Event> {
+  event(id: number): Promise<Event> {
     return this.eventRepository.findOne(id);
   }
 
-  findHostEvents(userId: number): Promise<Event[]> {
+  hostEvents(hostId: number): Promise<Event[]> {
     return this.eventRepository.find({
       relations: ['host'],
       where: {
         host: {
-          id: userId,
+          id: hostId,
         },
       },
     });
   }
 
-  findCoHostEvents(userId: number): Promise<Event[]> {
+  coHostEvents(coHostId: number): Promise<Event[]> {
     return this.eventRepository
       .createQueryBuilder('event')
       .leftJoin('event.coHosts', 'user')
-      .where('user.id = :userId', { userId })
+      .where('user.id = :userId', { userId: coHostId })
       .getMany();
   }
 
-  findJoinedEvents(userId: number): Promise<Event[]> {
+  participantEvents(userId: number): Promise<Event[]> {
     return this.eventRepository
       .createQueryBuilder('event')
       .leftJoin('event.participants', 'user')
@@ -51,7 +51,7 @@ export class EventService {
    * Find all the events that the user has not joined, created or is not co-hosting
    * @param userId
    */
-  findDiscoverEvent(userId: number): Promise<Event[]> {
+  discoverEvents(userId: number): Promise<Event[]> {
     return this.eventRepository
       .createQueryBuilder('event')
       .leftJoin('event.coHosts', 'coHost')
@@ -76,23 +76,29 @@ export class EventService {
       .getMany();
   }
 
-  async create(eventInput: CreateEventInput): Promise<Event> {
-    return await this.eventRepository.save(eventInput);
+  async createEvent(input: CreateEventInput): Promise<Event> {
+    return await this.eventRepository.save(input);
   }
 
-  async update(eventInput: UpdateEventInput): Promise<Event> {
-    await this.eventRepository.update(eventInput.id, { ...eventInput });
-    return this.findOne(eventInput.id);
+  async updateEvent(input: UpdateEventInput): Promise<Event> {
+    await this.eventRepository.update(input.id, { ...input });
+    return this.event(input.id);
   }
 
-  async addParticipant(eventId: number, userId: number) {
+  async deleteEvent(id: number): Promise<Event> {
+    const event = await this.event(id);
+    await this.eventRepository.delete(event.id);
+    return event;
+  }
+
+  async joinEvent(eventId: number, userId: number) {
     const user = new User(userId);
     const event = await this.eventRepository.findOne(eventId, { relations: ['participants'] });
     event.participants.push(user);
     return await this.eventRepository.save(event);
   }
 
-  async removeParticipant(eventId: number, userId: number) {
+  async leaveEvent(eventId: number, userId: number) {
     const event = await this.eventRepository.findOne(eventId, { relations: ['participants'] });
     if (event == undefined) {
       throw new Error('Event does not exist');
@@ -106,11 +112,5 @@ export class EventService {
       event.participants.splice(index, 1);
       return await this.eventRepository.save(event);
     }
-  }
-
-  async delete(id: number): Promise<Event> {
-    const event = await this.findOne(id);
-    await this.eventRepository.delete(event.id);
-    return event;
   }
 }
--- a/apps/mull-api/src/app/event/event.service.ts
+++ b/apps/mull-api/src/app/event/event.service.ts
@@ -12,15 +12,15 @@ export class EventService {
     private eventRepository: Repository<Event>
   ) {}
 
-  events(): Promise<Event[]> {
+  getAllEvents(): Promise<Event[]> {
     return this.eventRepository.find();
   }
 
-  event(id: number): Promise<Event> {
+  getEvent(id: number): Promise<Event> {
     return this.eventRepository.findOne(id);
   }
 
-  hostEvents(hostId: number): Promise<Event[]> {
+  getEventsHostedByUser(hostId: number): Promise<Event[]> {
     return this.eventRepository.find({
       relations: ['host'],
       where: {
@@ -31,7 +31,7 @@ export class EventService {
     });
   }
 
-  coHostEvents(coHostId: number): Promise<Event[]> {
+  getEventsCoHostedByUser(coHostId: number): Promise<Event[]> {
     return this.eventRepository
       .createQueryBuilder('event')
       .leftJoin('event.coHosts', 'user')
@@ -39,7 +39,7 @@ export class EventService {
       .getMany();
   }
 
-  participantEvents(userId: number): Promise<Event[]> {
+  getEventsAttendedByUser(userId: number): Promise<Event[]> {
     return this.eventRepository
       .createQueryBuilder('event')
       .leftJoin('event.participants', 'user')
@@ -51,7 +51,7 @@ export class EventService {
    * Find all the events that the user has not joined, created or is not co-hosting
    * @param userId
    */
-  discoverEvents(userId: number): Promise<Event[]> {
+  getEventsRecommendedToUser(userId: number): Promise<Event[]> {
     return this.eventRepository
       .createQueryBuilder('event')
       .leftJoin('event.coHosts', 'coHost')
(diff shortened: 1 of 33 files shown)


apps/mull-api/src/app/user/user.service.ts
```diff
@@ -12,11 +13,11 @@ export class UserService {
     private userRepository: Repository<User>
   ) {}
 
-  findAll(): Promise<User[]> {
+  users(): Promise<User[]> {
     return this.userRepository.find();
   }
 
-  findOne(id: number): Promise<User> {
+  user(id: number): Promise<User> {
```
r-saba: How do we feel about `findUser`? with just `user` I had to take a second to think and go see what the function did
```suggestion
  findUser(id: number): Promise<User> {
```

apps/mull-api/src/app/event/event.service.ts
```diff
@@ -12,34 +12,34 @@ export class EventService {
     private eventRepository: Repository<Event>
   ) {}
 
-  findAll(): Promise<Event[]> {
+  events(): Promise<Event[]> {
```
r-saba: I prefer `eventService.findAllEvents()`  vs `eventService.events()`. In the latter I have to take half a second to think about what it's doing

apps/mull-api/src/app/event/event.service.ts
```diff
@@ -12,34 +12,34 @@ export class EventService {
...
     return this.eventRepository.find();
   }
 
-  findOne(id: number): Promise<Event> {
+  event(id: number): Promise<Event> {
```
r-saba: Do we like this instead?
```suggestion
  findEvent(id: number): Promise<Event> {
```

apps/mull-api/src/app/event/event.service.ts
```diff
@@ -12,34 +12,34 @@ export class EventService {
...
     return this.eventRepository.findOne(id);
   }
 
-  findHostEvents(userId: number): Promise<Event[]> {
+  hostEvents(hostId: number): Promise<Event[]> {
```
r-saba: I was going to refactor this to one of the below. What ya think?
```suggestion
  getEventsHostedByUser(userId: number): Promise<Event[]> {
```
or
```suggestion
 eventsHostedBy(hostId: number): Promise<Event[]> {
```

r-saba: If we do change it to one of the above, we'd change `coHostEvents` and `participantEvents` as well below.

apps/mull-api/src/app/event/event.service.ts
```diff
@@ -51,7 +51,7 @@ export class EventService {
    * Find all the events that the user has not joined, created or is not co-hosting
    * @param userId
    */
-  findDiscoverEvent(userId: number): Promise<Event[]> {
+  discoverEvents(userId: number): Promise<Event[]> {
```
r-saba: How do we feel about this
```suggestion
  recommendedEvents(userId: number): Promise<Event[]> {
```

apps/mull-api/src/app/event/event.service.ts
```diff
@@ -76,23 +76,29 @@ export class EventService {
...
-    await this.eventRepository.update(eventInput.id, { ...eventInput });
-    return this.findOne(eventInput.id);
+  async deleteEvent(id: number): Promise<Event> {
+    const event = await this.event(id);
+    await this.eventRepository.delete(event.id);
+    return event;
   }
 
-  async addParticipant(eventId: number, userId: number) {
+  async joinEvent(eventId: number, userId: number) {
```
r-saba: I feel `eventService.addEventParticipant()` is more descriptive than `eventService.joinEvent()`
```suggestion
  async addEventParticipant(eventId: number, userId: number) {
```

r-saba: If we change this, we should change the below to `removeEventParticipant`

apps/mull-api/src/app/media/media.service.ts
```diff
@@ -13,6 +13,18 @@ export class MediaService {
...
+      await this.saveFile(file);
+      var media = await this.createMedia(file.mimetype);
+      this.updateFilename(file.filename, media.id, media.mediaType);
+    } catch (err) {
+      return new Error('Internal Server Error');
+    }
+
+    return media;
+  }
+
```
r-saba, line 27: very clean

apps/mull-api/src/app/user/user.service.ts
r-saba: If we change it here, let's change the below to `findFriends`

apps/mull-ui/src/app/pages/create-event/create-event.tsx
```diff
@@ -107,7 +107,7 @@ const CreateEventPage = ({ history }: CreateEventProps) => {
         } = await uploadFile({ variables: { file: file } });
         await createEvent({
           variables: {
-            createEventInput: {
+            input: {
```
r-saba: Not a big fan of this vague `input`, could we do `event` or `eventInput`?

apps/mull-ui/src/app/pages/event-page/event-page.tsx
```diff
@@ -26,8 +26,8 @@ export const EventPage = ({
   const { id } = useParams<{ id: string }>();
   const eventId = parseInt(id);
 
-  const { loading, error, data } = useFindSpecificEventQuery({
-    variables: { eventId },
+  const { loading, error, data } = useEventQuery({
```
r-saba, line 29:
```suggestion
  const { loading, error, data } = useFindEventQuery({
```

apps/mull-ui/src/app/pages/home/my-events/my-events.spec.tsx
```diff
@@ -27,8 +27,8 @@ describe('myEventsPage', () => {
       const mocks: MockedResponse[] = [
         {
           request: {
-            query: GetUsersEventsDocument,
-            variables: { UserId: 1 },
+            query: OwnedEventsDocument,
```
r-saba, line 30: I'm not familar with graphql codegen yet, why does it add "Document" at the end?

r-saba (changes requested):
Thanks for this. 
Like I mentioned in our slack conversation. I was thinking of doing this as well, so I had some names in mind already for the functions. Let me know how you feel about them.

I spent a lot of time reviewing the code, so didn't have time to look at the test files

apps/mull-ui/src/app/pages/home/my-events/my-events.spec.tsx
cristian-aldea: That's the raw query string, which is what we need to put here.

apps/mull-ui/src/app/pages/create-event/create-event.tsx
cristian-aldea: I find it redundant since you should already know what the query is. It's the input to the create event query, idk if we need to mention that it's the create event input or event input.

cristian-aldea:
Overall @r-saba I don't mind your suggestions, but I find that they might be a bit too verbose. My main concern was having identical names between the services and resolvers, not really the specific names themselves

apps/mull-api/src/app/media/media.service.ts
cristian-aldea: It was actually nice getting to look at that method and realize it could be simplified, made me more comfortable doing stuff in the backend.

apps/mull-api/src/app/user/user.service.ts
cristian-aldea: sure I'm good with `findUser`

apps/mull-api/src/app/event/event.service.ts
cristian-aldea: Maybe `getAllEvents`? You're not really searching or filtering, you just want all of them

apps/mull-api/src/app/event/event.service.ts
cristian-aldea: I like the second one, more concise.

apps/mull-ui/src/app/pages/create-event/create-event.tsx
cristian-aldea: If you insist that being explicit about it helps then I don't mind

apps/mull-ui/src/app/pages/event-page/event-page.tsx
cristian-aldea: No worries, if we change the names I'll take care of making sure all references are renamed properly.

apps/mull-api/src/app/event/event.service.ts
r-saba: `getAllEvents` sounds good as well

apps/mull-ui/src/app/pages/create-event/create-event.tsx
r-saba: I don't insist, if you'd prefer to keep it input it's okay too. I just felt it was easier to read

apps/mull-api/src/app/user/user.service.ts
r-saba: > Overall @r-saba I don't mind your suggestions, but I find that they might be a bit too verbose. My main concern was having identical names between the services and resolvers, not really the specific names themselves

Ahh I see, my concern was more on the names. I noticed teammates getting confused while using the graphql editor on what exactly the query or mutation did.

r-saba:
> Overall @r-saba I don't mind your suggestions, but I find that they might be a bit too verbose. My main concern was having identical names between the services and resolvers, not really the specific names themselves

Ahh I see, my concern was more on the names. I noticed teammates getting confused while using the graphql editor on what exactly the query or mutation did.

apps/mull-api/src/app/user/user.service.ts
cristian-aldea: I'm thinking of putting the name `getUser` for the service method name

apps/mull-api/src/app/event/event.service.ts
cristian-aldea: I'll be putting the name `getEvent` for the service method.

apps/mull-api/src/app/event/event.service.ts
cristian-aldea: I'll be putting your first suggestion as the service name, although the second is more concise it doesn't matter as much for the service method.

apps/mull-api/src/app/event/event.service.ts
cristian-aldea: That or `getRecommendedEvents`, what do you think?

apps/mull-api/src/app/event/event.service.ts
cristian-aldea: That seems like a great name! I'll use your suggestion

apps/mull-ui/src/app/pages/create-event/create-event.tsx
cristian-aldea: Maybe instead of `input`, `event` could be used? I'm not a fan of putting `createEventInput` or something similarly long, but `event` clearly shows that what you need to give as input is an event, or at least part of an event

apps/mull-api/src/app/event/event.service.ts
cristian-aldea: For `participantEvents`, what do you think about `getEventsAttendedByUser`?

image

apps/mull-api/src/app/event/event.service.ts
cristian-aldea: Oh, or `getEventsRecommendedToUser`, to make it match up better with the other methods

apps/mull-api/src/app/event/event.service.ts
r-saba: I like it. We could also do `eventsAttendedByUser`

r-saba: Very clear to me

apps/mull-ui/src/app/pages/create-event/create-event.tsx
r-saba: Sounds good to me

apps/mull-api/src/app/event/event.service.ts
r-saba: Yeah, that works if it better matches the other methods

cristian-aldea:
@r-saba I've implemented your changes and recommendations and ran all linting and tests and the app to make sure nothing 
🅱️roke, have another look when you can.

r-saba (approved):
Looks really nice!! Thanks for going the extra mile. LGTM!

bit172 (approved):
Method names look good to me. Good work!

PR #182, opened 2021-01-29
SONARCLOUD: Fix Code Coverage and Code Duplication

cindyslittleplanet:
This PR aims to increase the code coverage by ignoring files that were not meant to be counted in the code coverage and decrease the code duplication by removing some filler chat bubbles in announcements as they are only static placeholders. 

Before: 
Screen Shot 2021-01-29 at 11 48 12 AM

After: 
Screen Shot 2021-01-29 at 11 51 53 AM

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -13,7 +13,7 @@ sonar.links.scm=https://github.com/RGPosadas/Mull
 sonar.host.url=https://sonarcloud.io
 sonar.organization=rgposadas
 sonar.sources=libs, apps
-sonar.exclusions=**/*jest*.js,**/*spec*
+sonar.exclusions=**/*jest*.js,**/*spec*,**/mull-ui-e2e/**,**/database/**,**/environments/**,**/strategies/**, **/generated/**
 
 # Code coverage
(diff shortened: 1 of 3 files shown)


sonar-project.properties
```diff
@@ -13,7 +13,7 @@ sonar.links.scm=https://github.com/RGPosadas/Mull
 sonar.host.url=https://sonarcloud.io
 sonar.organization=rgposadas
 sonar.sources=libs, apps
-sonar.exclusions=**/*jest*.js,**/*spec*
+sonar.exclusions=**/*jest*.js,**/*spec*,**/mull-ui-e2e/**,**/database/**,**/environments/**,**/strategies/**, **/generated/**
```
cristian-aldea, line 16: Idk if `strategies` should be excluded.

cristian-aldea (approved):
Good work, just had a comment

sonar-project.properties
cindyslittleplanet: The reason why I excluded `strategies` was because those were the Facebook, Twitter and Google redirects for login, so I don't think it's in our scope to cover those part with tests.

r-saba (approved):
LGTM! Announcements has a lot of code duplication which will be removed once the corresponding User Story is completed.

PR #154, opened 2021-01-16
US-2.8: Joining events

leojrsilao:
This PR closes #66 

To properly test this PR you will need to set up your database such that User with id 1: 
- Is not involved in at least 1 event
 
**Steps to test joining events**
1. Make sure to run `npm i ` on the project to pickup the new dependencies 
2. Run `npm run gen-graphql` to generate the `graphql.tsx`
1. Run `nx serve mull-api` and `nx serve mull-ui`
2. Go to `http://localhost:4200/home/discover`
3. Join an event
3.a By clicking on the join event button on the event card
3.b By clicking on the event card, it brings you to the preview page for the event. Click on "Join" button on the event preview page 

**Steps to run the E2E tests:**
1. Make sure to run `npm i ` on the project to pickup the new dependencies 
1. Run the server `nx serve mull-api`
1. In the root directory run  `nx e2e mull-ui-e2e --watch` ( Cypress takes some time to open up)

PR #183, opened 2021-01-30
FIX: Convert raw SQL to TypeORM syntax

r-saba:
This PR removes technical debt introduced in #154 by converting the raw SQL in `event.service.ts` into typeORM syntax.

Having all our queries use the ORM syntax will make our backend easier to maintain and update as new features are added.

I had to update a failing snapshot from develop. It is unrelated to this PR.

*Notes: Discover page only shows public events. This means events with `restriction = 0`

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,12 @@ To test all projects:
 nx run-many --target=test --all
 ```
 
+To test a specific file:
+
+```bash
+nx test <project> --test-file <file-name>
+```
+
 To test projects affected by your changes:
 
 ```bash
--- a/apps/mull-api/src/app/event/event.service.spec.ts
+++ b/apps/mull-api/src/app/event/event.service.spec.ts
@@ -4,12 +4,7 @@ import { Repository } from 'typeorm';
 import { Event } from '../entities';
 import { mockAllUsers } from '../user/user.mockdata';
 import { MockType } from '../user/user.service.spec';
-import {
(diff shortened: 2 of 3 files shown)


ys-lin (commented):
I ran the tests and there are some test failures...
image

leojrsilao (approved):
LGTM!! 🚀 🚀  The query looks cleaner now and makes the code more consistent with our other queries.

cristian-aldea (approved):

cristian-aldea:
Oh one thing I would like to mention is that I think you should revert your changes to the snapshot, since it currently doesn't work for everyone as you outlined in #195, no need to change it.

cristian-aldea (changes requested):
Can you first run snapshots tests on my PR #196 and let me know if it passes. If it does, please revert your snapshot changes. If not I'll reapprove

r-saba:
> Can you first run snapshots tests on my PR #196 and let me know if it passes. If it does, please revert your snapshot changes. If not I'll reapprove

I reverted the snapshot commit.

README.md
```diff
@@ -128,6 +128,12 @@ To test all projects:
 nx run-many --target=test --all
 ```
 
+To test a specific file:
```
cristian-aldea, line 131: 👍

cristian-aldea (approved):

PR #194, opened 2021-02-01
FIX: Add Location Seeding

### LINKED ISSUE #198: FIX: Add Location Seeding
Locations are now expected to be returned when querying events in the database. However, seeding doesn't create any Locations for events, which makes it so that the application breaks when seeding the database.

osahtout:
This is just a quick fix to add location to the seeder. 

Also removed some images type in the media seeder for a more controlled media seeding.

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,8 +232,14 @@ Prerequisite:
 seeding:
 
 1. Start the database, if running locally
-1. Run `npm run start mull-api` to update the columns (You don't have to have this running afte updating)
+1. Run `npm run start mull-api` to update the columns (You don't have to have this running after updating)
 1. Run `npm run seed:run`
+
+For images to show on the app after seeding:
+
+Download the Images from [the image folder in the drive](https://drive.google.com/drive/folders/1Czg2p3kZMvOyvrGpTfBH5bSq4sJVQZsJ?usp=sharing) and add them to the upload folder in mull-api.
+
+This only needs to be done once and other images can be added.
 
 ## Travis CI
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -229,6 +229,12 @@ seeding:
 1. Run `npm run start mull-api` to update the columns (You don't have to have this running afte updating)
 1. Run `npm run seed:run`
 
+For images to show on the app after seeding:
+
+Download the Images from [the image folder in the drive](https://drive.google.com/drive/folders/1Czg2p3kZMvOyvrGpTfBH5bSq4sJVQZsJ?usp=sharing) and add them to the upload folder in mull-api.
+
+This only needs to be done once and other images can be added/
```
cristian-aldea:
```suggestion
This only needs to be done once and other images can be added.
```

cristian-aldea (approved):

README.md
```diff
@@ -229,6 +229,12 @@ seeding:
 1. Run `npm run start mull-api` to update the columns (You don't have to have this running afte updating)
```
r-saba:
```suggestion
1. Run `npm run start mull-api` to update the columns (You don't have to have this running after updating)
```

r-saba (commented):
I noticed this is only populating the `title` column of the location table. Other than showing up on the front-end this isn't very useful for developers who will need to work with placeId's and the `Point` table. Can we populate these as well or is there a reason we're not doing this?

osahtout:
> 
> 
> I noticed this is only populating the `title` column of the location table. Other than showing up on the front-end this isn't very useful for developers who will need to work with placeId's and the `Point` table. Can we populate these as well or is there a reason we're not doing this?

I can definitely try, it may be a little hacky

JimLau49 (approved):
LGTM

r-saba (approved):
Approved but please make a ticket for the placeID seeding.

PR #200, opened 2021-02-05
TASK-45: Authenticated Backend Routes

### LINKED ISSUE #169: TASK-45: Setup Authenticated Backend Routes
**Task Description**

As a developer, I want authenticated backend routes so that some endpoints are protected.

**Acceptance Criteria**

- [ ] Authenticated routes can only be accessed by authenticated users
- [ ] Have a reusable Guard in the backend

r-saba:
Closes #169 

Our Routes are secured by either using the newly created `AuthGuard` or the `AuthenticatedUser` class and function in `auth.guard.ts`.

Using  `AuthenticatedUser` we can get the userId directly from the header of the request. This means we no longer should be sending userId as a parameter to any graphQL queries or mutations.

For queries or mutations not taking userId as a parameter, I created the `AuthGuard` class. When placed on an endpoint it checks that the header has a valid access token and continues with the request. If the access token is invalid an error is thrown and the request is not processed.

**Notes**
Queries to our GraphQL server now requires a header with the following field
`Auhtorization: 'Bearer <valid-accesToken>'`

This is what it would look like in apollo studio
image

Ideally we'd have a script to generate access tokens, I haven't found a pretty solution I'm comfortable pushing at the moment. Feel free to work on this if it interests you.

You can get a valid access token by going to the network tab in your browser and looking at the header of a graphql query made when logged in.

image

**Running e2e tests**
You will need to add a  `cypress.env.json` file in mull-ui-e2e. Take a look at the docs and resources channel to see what goes inside

### DIFF
--- a/apps/mull-api/src/app/auth/auth.guard.spec.ts
+++ b/apps/mull-api/src/app/auth/auth.guard.spec.ts
@@ -1,7 +1,9 @@
 import { createMock } from '@golevelup/nestjs-testing';
 import { ExecutionContext } from '@nestjs/common';
+import { ROUTE_ARGS_METADATA } from '@nestjs/common/constants';
 import jwt from 'jsonwebtoken';
 import { AuthenticatedUser, AuthGuard } from './auth.guard';
+
 jest.mock('jsonwebtoken');
 const mockedJwt = jwt as jest.Mocked<typeof jwt>;
 
@@ -16,7 +18,7 @@ describe('AuthGuards', () => {
       expect(guard).toBeDefined();
     });
 
-    it('Guard should return true with auth', () => {
+    it('should return true with auth', () => {
       const context = createMock<ExecutionContext>();
 
       context.switchToHttp().getNext.mockReturnValue({
@@ -31,26 +33,48 @@ describe('AuthGuards', () => {
       expect(guard.canActivate(context)).toBeTruthy();
     });
 
-    it('Guard should return false without auth', () => {
+    it('should return false with invalid auth', () => {
       const context = createMock<ExecutionContext>();
 
       context.switchToHttp().getNext.mockReturnValue({
         req: {
           headers: {
-            authorization: 'auth',
+            authorization: '',
           },
         },
       });
 
       mockedJwt.verify.mockImplementation(() => false);
-      expect(guard.canActivate(context)).toBeFalsy();
+      expect(guard.canActivate(context)).toBe(false);
+    });
+
+    it('should throw exception with invalid header', () => {
+      const context = createMock<ExecutionContext>();
+
+      context.switchToHttp().getNext.mockReturnValue({
+        req: {
+          headers: {},
+        },
+      });
+
+      mockedJwt.verify.mockImplementation(() => false);
+      expect(() => guard.canActivate(context)).toThrow('Unauthorized');
     });
   });
-
+  // This testing methodology for param decorators was found on the nestjs github repo
+  // https://github.com/nestjs/nest/issues/1020
   describe('Auth Param Decorator', () => {
-    it('Auth param decorator should return with auth', () => {
-      const context = createMock<ExecutionContext>();
+    function getParamDecoratorFactory(decorator: Function) {
+      class Test {
+        public test(@decorator() value) {}
+      }
 
+      const args = Reflect.getMetadata(ROUTE_ARGS_METADATA, Test, 'test');
+      return args[Object.keys(args)[0]].factory;
+    }
+
+    it('should return userId with auth', () => {
+      const context = createMock<ExecutionContext>();
       context.switchToHttp().getNext.mockReturnValue({
         req: {
           headers: {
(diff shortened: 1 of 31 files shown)


apps/mull-api/src/app/auth/auth.guard.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+}
+
+export const AuthenticatedUser = createParamDecorator((_, context: ExecutionContext): number => {
+  const accessToken = context.switchToHttp().getNext().req.headers.authorization.split(' ')[1];
+  try {
+    const { id } = jwt.verify(accessToken, environment.jwt.accessSecret) as IAuthToken;
+    return id;
+  } catch (err) {
+    throw new UnauthorizedException('Unauthorized');
```
bit172: Nice and clean. I like it.  🚀

apps/mull-api/src/app/auth/auth.guard.ts
```diff
@@ -0,0 +1,33 @@
...
+  UnauthorizedException,
+} from '@nestjs/common';
+import jwt from 'jsonwebtoken';
+import { Observable } from 'rxjs';
+import { environment } from '../../environments/environment';
+
+@Injectable()
+export class AuthGuard implements CanActivate {
+  canActivate(context: ExecutionContext): boolean | Promise<boolean> | Observable<boolean> {
+    const accessToken = context.switchToHttp().getNext().req.headers.authorization.split(' ')[1];
```
bit172: I think you should check if the authorization header is present or bring it in the try block since the response returned is a 500 and not a 401 if it's not there. 
image

bit172 (changes requested):
Good work! 🚀 Just a small thing about the guard that I saw, the rest is good. Now that we extract the user Id from the access token, is having the user Id on the frontend obsolete? 🤔

apps/mull-api/src/app/auth/auth.guard.ts
r-saba: Good point! Moved it into the try block so we can control what error gets thrown.

r-saba:
> Good work! 🚀 Just a small thing about the guard that I saw, the rest is good. Now that we extract the user Id from the access token, is having the user Id on the frontend obsolete? 🤔

Hopefully yes, but I see it's still used in privateRoutes

bit172 (approved):
LGTM!
- [x] Authenticated routes can only be accessed by authenticated users
- [x] Have a reusable Guard in the backend

package-lock.json
```diff
@@ -10328,22 +10334,14 @@
           }
         },
         "graphql-upload": {
-          "version": "11.0.0",
-          "resolved": "https://registry.npmjs.org/graphql-upload/-/graphql-upload-11.0.0.tgz",
-          "integrity": "sha512-zsrDtu5gCbQFDWsNa5bMB4nf1LpKX9KDgh+f8oL1288ijV4RxeckhVozAjqjXAfRpxOHD1xOESsh6zq8SjdgjA==",
+          "version": "8.1.0",
```
RGPosadas: Why did we downgrade graphql-upload + remove other deps? When I run `npm i`, it brings graphql-upload  back to 11.0.0, and brings back `fs-capacitor` and `isobject`

apps/mull-api/src/app/user/user.resolver.ts
```diff
@@ -24,17 +27,19 @@ export class UserResolver {
   }
 
   @Mutation(/* istanbul ignore next */ () => User)
+  @UseGuards(AuthGuard)
```
RGPosadas: I don't think we need it here for createUser as I just tried to create an account and it told me I was unauthorized.

RGPosadas (changes requested):
Locally, everything works great after taking out the guard on createUser!

However, blocking due to 4 failing mull-ui unit tests (from the pages that you've modified), and the e2e tests are broken as well due to the guards. Example from US-1.1, it fails to create the event with a Fatal Error:
image

leojrsilao (changes requested):
Everything looks great with the guards 🚀. 
Noticed as well that there are 4 unit tests failing in the front end and some e2e tests are failing because of the guards.

package-lock.json
r-saba: Good question I actually have no idea why, ran `npm i` again and it upgraded to the correct version

r-saba: Done

apps/mull-api/src/app/user/user.resolver.ts
r-saba: Done

apps/mull-api/src/app/event/inputs/event.input.ts
```diff
@@ -3,12 +3,11 @@ import { Field, InputType, Int } from '@nestjs/graphql';
 import { Event, Location, Media } from '../../entities';
 import { LocationInput } from '../../location/inputs/location.input';
 import { MediaInput } from '../../media/inputs/media.input';
-import { UserInput } from '../../user/inputs/user.input';
 
 @InputType()
 export class CreateEventInput {
   @Field()
-  title: string;
+  title?: string;
```
RGPosadas: Why optional title?

RGPosadas (commented):
Great work! Tests + linting also pass.
Though please run the backend to also update mull-api's schema.gql as there is still a remnant of hostId.

package-lock.json
cristian-aldea: I think i know why, when you do `npm i package` it probably doesn't force resolutions. I ran `npm i` again too and then it finally force resolved my packages.

package.json
```diff
@@ -96,6 +96,7 @@
     "@babel/preset-env": "7.9.6",
     "@babel/preset-react": "7.9.4",
     "@babel/preset-typescript": "7.9.0",
+    "@golevelup/nestjs-testing": "^0.1.2",
```
cristian-aldea, line 100: ❔ Why was this package added? I see that its used to mock the execution context for nestjs. But I'm wondering if its possible to mock it without a new package.

apps/mull-api/src/app/auth/auth.guard.spec.ts
```diff
@@ -0,0 +1,81 @@
+import { createMock } from '@golevelup/nestjs-testing';
```
cristian-aldea, line 1: 💡I see that some code still isn't being covered. Might be good to test those lines

image

cristian-aldea (commented):
Looks good, currently, the application works just fine, and testing and lining are passing. I can't run the e2e tests thou for some reason:
image


Just leaving comments for now

cindyslittleplanet (approved):
Everything passes now, so LGTM! 

- [x] Authenticated routes can only be accessed by authenticated users
- [x] Have a reusable Guard in the backend

apps/mull-api/src/app/auth/auth.guard.spec.ts
r-saba: Assumed since I was testing the function I would get coverage. I found this [github issue](https://github.com/nestjs/nest/issues/1020) and after applying it, it gives us our line coverage while testing the function too.

package.json
r-saba: It's a nice testing utility for our project. Especially for mocking context. Here's a snippet from their docs

> This function will create a mock object for you with all sub properties mocked as jest.fn() unless otherwise provided, to allow for easy mocking later on, but more on that later.

I also use this repo as a reference for nestjs testing and that's where I found this utility.  [testing-nestjs](https://github.com/jmcdo29/testing-nestjs)

apps/mull-api/src/app/event/inputs/event.input.ts
r-saba: Shouldn't be, done

RGPosadas (approved):
Thanks for addressing all my concerns! LGTM 🚀 

**Acceptance Criteria**

- [x] Authenticated routes can only be accessed by authenticated users
- [x] Have a reusable Guard in the backend

leojrsilao (approved):
LGTM! Addressed my concerns 👍 🚀 

**Acceptance Criteria**

- [x] Authenticated routes can only be accessed by authenticated users
- [x] Have a reusable Guard in the backend

bit172 (approved):

PR #212, opened 2021-02-13
TASK-50: Save Messages to back end

### LINKED ISSUE #184: TASK-50: Save Messages to Back-end
**Task Description**

As a developer, I want to save messages to the backend so that they can be fetched from the frontend.

**Acceptance Criteria**
- [ ] Is reusable for announcements and group chats
- [ ] Messages are saved when they are created
- [ ] Messages can include images

JimLau49:
This PR implements #184 

To test:
1. `nx serve mull-api`
2. pop into `localhost:3333/graphql`
3. Run the `postAdded` subscriber first by following the docs section in the gql playground or see the attached screenshot
image

       You should see `listening...` in the output 
      image

4. In a new tab, create a post using `createPost` mutation (see below for an example)
image
5. Go back to the postAdded subscriber to see the message appear

Expected result:
image

### DIFF
--- a/apps/mull-api/src/app/channel/inputs/channel.input.ts
+++ b/apps/mull-api/src/app/channel/inputs/channel.input.ts
@@ -1,0 +1,8 @@
+import { Field, InputType } from '@nestjs/graphql';
+import { Channel } from '../../entities';
+
+@InputType()
+export class ChannelInput implements Partial<Channel> {
+  @Field()
+  id: number;
+}
--- a/apps/mull-api/src/app/entities/post.entity.ts
+++ b/apps/mull-api/src/app/entities/post.entity.ts
@@ -1,3 +1,4 @@
+import { Field, Int, ObjectType } from '@nestjs/graphql';
 import {
   Column,
   Entity,
@@ -13,30 +14,38 @@ import { PostReaction } from './post-reaction.entity';
 import { User } from './user.entity';
 
 @Entity()
+@ObjectType()
 export class Post {
+  @Field(/* istanbul ignore next */ () => Int)
   @PrimaryGeneratedColumn()
   id: number;
 
-  @OneToOne(() => User)
-  @JoinColumn()
+  @Field(/* istanbul ignore next */ () => User)
+  @ManyToOne(/* istanbul ignore next */ () => User, /* istanbul ignore next */ (user) => user.posts)
   user: User;
 
-  @OneToOne(() => Post)
+  @Field({ nullable: true })
+  @OneToOne(/* istanbul ignore next */ () => Post)
   @JoinColumn()
-  parentPost: Post;
+  parentPost?: Post;
 
+  @Field()
   @Column()
   message: string;
 
+  @Field()
   @Column()
   createdTime: Date;
 
+  @Field(/* istanbul ignore next */ () => [Media], { nullable: true })
   @OneToMany(() => Media, (media) => media.post)
-  medias: Media[];
+  medias?: Media[];
 
-  @OneToMany(() => PostReaction, (reaction) => reaction.post)
-  reactions: PostReaction[];
+  @Field(/* istanbul ignore next */ () => [PostReaction], { nullable: true })
+  @OneToMany(/* istanbul ignore next */ () => PostReaction, (reaction) => reaction.post)
+  reactions?: PostReaction[];
 
-  @ManyToOne(() => Channel, (channel) => channel.posts)
+  @Field(/* istanbul ignore next */ () => Channel)
+  @ManyToOne(/* istanbul ignore next */ () => Channel, (channel) => channel.posts)
   channel: Channel;
 }
--- a/apps/mull-api/src/app/post/inputs/post-reaction.input.ts
+++ b/apps/mull-api/src/app/post/inputs/post-reaction.input.ts
@@ -1,0 +1,8 @@
+import { Field, InputType, Int } from '@nestjs/graphql';
+import { PostReaction } from '../../entities';
+
+@InputType()
+export class PostReactionInput implements Partial<PostReaction> {
+  @Field(/* istanbul ignore next */ () => Int)
+  id: number;
+}
--- a/apps/mull-api/src/app/post/post.resolver.ts
+++ b/apps/mull-api/src/app/post/post.resolver.ts
@@ -1,0 +1,41 @@
+import { Args, Int, Mutation, Query, Resolver, Subscription } from '@nestjs/graphql';
+import { PubSub } from 'graphql-subscriptions';
+import { Post } from '../entities';
+import { CreatePostInput, UpdatePostInput } from './inputs/post.input';
+import { PostService } from './post.service';
+
+const pubSub = new PubSub();
+
+@Resolver(() => Post)
+export class PostResolver {
+  constructor(private readonly postService: PostService) {}
+
+  @Subscription(() => Post)
+  postAdded() {
+    return pubSub.asyncIterator(['postAdded']);
+  }
+
+  @Mutation(/* istanbul ignore next */ () => Post)
+  async post(@Args('post') post: CreatePostInput) {
+    const savedPost = this.postService.createPost(post);
+    pubSub.publish('postAdded', {
+      postAdded: savedPost,
+    });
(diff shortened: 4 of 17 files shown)


apps/mull-api/src/schema.gql
```diff
@@ -75,16 +98,28 @@ type Mutation {
   createLocation(location: LocationInput!): Location!
   createUser(user: CreateUserInput!): User!
   deleteEvent(id: Int!): Event!
+  deletePost(id: Int!): Post!
   deleteUser: User!
   joinEvent(eventId: Int!): Boolean!
   leaveEvent(eventId: Int!): Boolean!
   login(loginInput: LoginInput!): LoginResult!
+<<<<<<< HEAD
```
RGPosadas: It seems you accidentally committed this merge conflict resolver. Since it's autogenerated, just delete this file and commit the updated schema

apps/mull-api/src/app/post/post.resolver.ts
```diff
@@ -0,0 +1,37 @@
...
+  async post(@Args('post') post: CreatePostInput) {
+    pubSub.publish('postAdded', {
+      postAdded: post,
+    });
+    return this.postService.createPost(post);
+  }
+
+  @Query(/* istanbul ignore next */ () => [Post])
+  async posts() {
+    return this.postService.getAllPosts();
```
RGPosadas, line 30: This line isn't covered/tested, and I think you can easily add the test for this.

RGPosadas (commented):
1. Please also run linting and remove the unused variables that the linter found.

2. Also saw that Jest thinks that there are uncovered lines in the new entities that you've added, but it's as always due to the decorators and not actual code that needs to be tested. I don't think we should add `/*istanbul ignore next*/` in all of those any longer, instead I think we should just let Jest ignore the `entities/` folder. What do you think?
image

3. Please also don't forget to add authentication to your `post.resolver` through AuthGuards!

apps/mull-api/src/app/channel/inputs/channel.input.ts
```diff
@@ -0,0 +1,8 @@
+import { Field, InputType } from '@nestjs/graphql';
+import { Channel } from '../../entities';
+
+@InputType()
+export class ChannelInputs implements Partial<Channel> {
```
cristian-aldea: 💡Why is this in plural form?
```suggestion
export class ChannelInput implements Partial<Channel> {
```

apps/mull-api/src/app/entities/post.entity.ts
```diff
@@ -13,30 +14,38 @@ import { PostReaction } from './post-reaction.entity';
...
 @Entity()
+@ObjectType()
 export class Post {
+  @Field(() => Int)
   @PrimaryGeneratedColumn()
   id: number;
 
-  @OneToOne(() => User)
-  @JoinColumn()
+  @Field(/* istanbul ignore next */ () => User)
```
cristian-aldea, line 23: 💡 You added `/* istanbul ignore next */` to these two lines, but not to all the other places where you added anon function in the entities. Might be good to also add it everywhere else.

apps/mull-api/src/app/entities/post.entity.ts
```diff
@@ -13,30 +14,38 @@ import { PostReaction } from './post-reaction.entity';
...
 
+  @Field(() => [Media], { nullable: true })
   @OneToMany(() => Media, (media) => media.post)
-  medias: Media[];
+  medias?: Media[];
 
+  @Field(() => [PostReaction], { nullable: true })
   @OneToMany(() => PostReaction, (reaction) => reaction.post)
-  reactions: PostReaction[];
+  reactions?: PostReaction[];
```
cristian-aldea: ⚠ Idk if these should be nullable since it should just return empty arrays when there's no media or reaction.

Did you run into TS issues, which made you put these as optional/nullable?

apps/mull-api/src/app/post/inputs/post-reaction.input.ts
```diff
@@ -0,0 +1,8 @@
+import { Field, InputType, Int } from '@nestjs/graphql';
+import { PostReaction } from '../../entities';
+
+@InputType()
+export class PostReactionInput implements Partial<PostReaction> {
+  @Field(/* istanbul ignore next */ () => Int)
+  id: number;
+}
```
cristian-aldea, line 8: 💡 I don't think this needs to be in a separate file, it could be moved to `post.input.ts`

apps/mull-api/src/app/post/post.resolver.ts
```diff
@@ -0,0 +1,37 @@
...
+  postAdded() {
+    return pubSub.asyncIterator(['postAdded']);
+  }
+
+  @Mutation(/* istanbul ignore next */ () => Post)
+  async post(@Args('post') post: CreatePostInput) {
+    pubSub.publish('postAdded', {
+      postAdded: post,
+    });
+    return this.postService.createPost(post);
```
cristian-aldea: ⚠⚠⚠ As discussed on slack, I think sending a notification to the subscriber before the post was saved successfully can be problematic:
- We don't get an id, since post here is a `CreatePostInput`. Ritchelle was confused as to why she couldn't get the id of the new post
- What if the post doesn't save successfully? Then the sub got a false notification since the backend effectively didn't change.

cristian-aldea (changes requested):
Looks great! I've left and few 💡suggestions and blocked the PR over some ⚠concerns

apps/mull-api/src/schema.gql
cristian-aldea: ⚠⚠⚠ I'm also blocking over this

apps/mull-api/src/app/post/inputs/post-reaction.input.ts
osahtout: Since it has a separate entity file I thought I would keep it consistant

apps/mull-api/src/app/entities/post.entity.ts
osahtout: It wouldn't let me create a post without creating media. Maybe I'm doing something wrong.

I think we should have a discussion about reactions.

osahtout: Though the channelInput is only nullable for this ticket. I will add a TODO

apps/mull-api/src/schema.gql
JimLau49: my bad will fix (idk how this keeps happening)

apps/mull-api/src/app/post/post.resolver.ts
osahtout: Good catch, I'll push a quick fix

apps/mull-api/src/app/entities/post.entity.ts
cristian-aldea: I actually should not have set it as blocking, its minor imo, but just wanted to know more about the why behind this change. Yeah lets discuss

cristian-aldea (approved):
Great work, thanks for addressing my comments

- [x] Is reusable for announcements and group chats
- [x] Messages are saved when they are created
- [x] Messages can include images

apps/mull-api/src/app/post/post.resolver.ts
JimLau49: Done

leojrsilao (approved):
LGTM! 🚀 

- [x] Is reusable for announcements and group chats
- [x] Messages are saved when they are created
- [x] Messages can include images

JimLau49:
> 1. Please also run linting and remove the unused variables that the linter found.
> 2. Also saw that Jest thinks that there are uncovered lines in the new entities that you've added, but it's as always due to the decorators and not actual code that needs to be tested. I don't think we should add `/*istanbul ignore next*/` in all of those any longer, instead I think we should just let Jest ignore the `entities/` folder. What do you think?
>    image
> 3. Please also don't forget to add authentication to your `post.resolver` through AuthGuards!

1. Done
2. We agree this should be done but in a separate ticket or fix PR
3.  Added AuthGuards to mutations and querys but subscription requires additional work that will be handled in #215

PR #213, opened 2021-02-13
BUG-12: Fix event start time that can come after the event end time during event creation

### LINKED ISSUE #163: BUG-12: Event Time Ends Before Start
**The Bug**

Users can enter an 'End Time' that is before a 'Start Time' even when the event is on the same day and still create the event.

**Steps To Reproduce**

1. Create a new event.
2. Make sure the event starts and ends the same day.
3. Put the 'Start Time' anytime after noon and put the 'End Time' anytime before noon.

Note: you can set any time you want, it doesn't have to be afternoon or forenoon. As long as the 'End TIme' is before the 'Start Time'.

**Expected Behavior**

The user should not be able to create the event either through UI disabling the times before 'Start Time' forcing the user to pick a time after the 'Start TIme' or through validation. 

**Bug Handling Strategy**

- [ ] Fix it now
- [X] Fix it later

ys-lin:
closes #163
To prevent users from creating an event that has a start time after the end time. 
image

### DIFF
--- a/apps/mull-ui-e2e/src/integration/US-1.1.spec.ts
+++ b/apps/mull-ui-e2e/src/integration/US-1.1.spec.ts
@@ -91,7 +91,6 @@ frameSizes.forEach((frame) => {
       cy.get('#imageFile').attachFile('../fixtures/trashed-park.jpg');
 
       cy.get('#startTime').type('11:20');
-
       cy.get('#endTime').type('08:20');
 
       cy.get('#eventTitle').type('test title');
@@ -102,18 +101,11 @@ frameSizes.forEach((frame) => {
       cy.get('#description').type('test description');
 
       cy.get('#location').click();
-      cy.get('#location-input-field').should('be.visible');
-
-      cy.get('#location-input-field').type('845 Rue Sherbrooke');
-      cy.get('#location-input-field-option-0', { timeout: 3500 }).should(
-        'contain.text',
-        '845 Rue Sherbrooke'
-      );
 
       cy.get('#location-input-field-option-0', { timeout: 5000 }).click();
       cy.get('[data-testid=pill-id-1]').click();
+
       cy.get('.create-event-button').click();
-      cy.get('.event-page-button').click();
 
       cy.get('#endTime ~ .error-message').should(
         'have.text',
--- a/apps/mull-ui-e2e/src/integration/US-1.1.spec.ts
+++ b/apps/mull-ui-e2e/src/integration/US-1.1.spec.ts
@@ -87,7 +87,7 @@ frameSizes.forEach((frame) => {
       cy.get('#location ~ .error-message').should('have.text', 'Event Location is required.');
     });
 
-    it('should show an error when end time is greater than start time', () => {
+    it('should show an error when end time is greater than start time if the event is on the same day', () => {
       cy.get('#imageFile').attachFile('../fixtures/trashed-park.jpg');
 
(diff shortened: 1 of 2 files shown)


JimLau49 (approved):
LGTM , nice use of moment :)

apps/mull-ui/src/app/pages/create-event/create-event.tsx
```diff
@@ -88,7 +89,12 @@ const CreateEventPage = ({ history }: CreateEventProps) => {
           return diff <= 30 * DAY_IN_MILLISECONDS;
         }),
       startTime: Yup.string().required('Start Time is required.'),
-      endTime: Yup.string().required('End Time is required.'),
+      endTime: Yup.string()
+        .required('End Time is required.')
+        .test('is-greater', 'The end time must be after the start time.', function (endTime) {
+          const { startTime } = this.parent;
+          return moment(endTime, 'HH:mm').isSameOrAfter(moment(startTime, 'HH:mm'));
+        }),
```
cristian-aldea: This is valid if the event is one day long. However, if the event takes place over multiple days then it might, for example, start at 5 pm on one day and finish at 10 am the next.

I'm not sure if its possible, but I would try to combine (startTime + startDate) => start and (endTime + endDate) => end and then comparing start and end.

apps/mull-ui-e2e/src/integration/US-1.1.spec.ts
```diff
@@ -87,6 +87,32 @@ frameSizes.forEach((frame) => {
       cy.get('#location ~ .error-message').should('have.text', 'Event Location is required.');
     });
 
+    it('should show an error when end time is greater than start time', () => {
```
cristian-aldea: There seems to be a linting issue with the e2e for US-2.7, I know you didn't cause it, but if you can it would be good to fix it.

cristian-aldea (commented):
Looks good, had a concern about validation with multi-day events, but I won't be blocking the PR over it, since I didn't have a suggestion for you.

apps/mull-ui/src/app/pages/create-event/create-event.tsx
osahtout: Or maybe call this function only when startDate == endDate?

cristian-aldea: true, that's a lot simpler

ys-lin: ah true, I'm going to fix this, nice catch 👍

apps/mull-ui-e2e/src/integration/US-1.1.spec.ts
ys-lin: Alright I will fix this

ys-lin: For the lint error, I'm not sure if it's worth fixing, since it's complaining about the "wait(200)" for being arbitrary.  I think Ritchelle put that to make the test less flaky. What I understand is that it waits 200ms after clicking the button before the assertions, and we might need to do some stuff with `cy.route `  and wait for things to load to make it not arbitrary, and that might be complicated.

cristian-aldea: no worries, was just a suggestion

cristian-aldea (approved):

PR #218, opened 2021-02-19
BUG-14: Fix Event Page Layout

### LINKED ISSUE #174: BUG-14: Entire Event Page Scrollable iOS
**Describe The Bug**

The entire event page, excluding the nav bars, is scrollable when only the second half should be. 

**To Reproduce**

On any iOS device, click on any event and use the event picture to scroll.

**Expected Behavior**

Nothing should happen. only when attempting to scroll from the second half where scrolling is allowed.

**Bug Handling Strategy**
- [ ] Fix it now
- [x] Fix it later

**Screenshots**

I Will add a screenshot at a later date

**Desktop (please complete the following information):**

N/A

**Smartphone (please complete the following information):**

- Tested on iOS devices
- May be in android devices (not tested)

cristian-aldea:
Initially, the event page implemented with a static layout where the header was part of the page flow. This caused issues on iOS and generally made things a bit more complicated. This pr aims to fix that by setting the header to be in a fixed position.
closes #174 

Notes:
- Took the change to refactor event-page-header and event-page-info out of the components folder and into the event-page folder, which is the only page where they are used.
- Added SCSS Intellisense as a recommended extension, I found it pretty useful


RGPosadas (changes requested):
Awesome, you fixed the event header! However, the event-info is spilling into it when scrolling up.
https://user-images.githubusercontent.com/33991121/108540352-14e00100-72af-11eb-9cea-19c0cd91f03f.mov

apps/mull-ui/src/app/pages/event-page/event-page-header/event-page-header.scss
```diff
@@ -1,4 +1,6 @@
-.event-page-header {
+@import '../../../../variables.scss';
+
+.event-page-header-container {
   padding: 0.5rem 0.5rem 0;
   box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.1);
 }
```
cindyslittleplanet, line 6: Adding the z-index will fix the problem of the event-info spilling up

```suggestion
.event-page-header-container {
  padding: 0.5rem 0.5rem 0;
  box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.1);
  z-index: 1;
}
```

cindyslittleplanet (commented):
On both mobile and web, the event date and time also doesn't fit with the event page header for me weirdly enough.
Otherwise, the event header stays in place on iOS now! 
image

apps/mull-ui/src/app/pages/event-page/event-page-header/event-page-header.scss
cristian-aldea: For some reason, I removed that when experimenting and forgot to put it back, my bad. As for the header content spilling over, I'm working on a fix for that.

cindyslittleplanet (approved):
The header stays in place now on iOS, so LGTM! 🚀 Good job!

apps/mull-ui-e2e/src/integration/US-2.7.spec.ts
```diff
@@ -2,6 +2,7 @@ import 'cypress-file-upload';
 import { geolocationStub } from '../fixtures';
 import { frameSizes } from './../fixtures/frame-sizes';
 
+/* eslint-disable cypress/no-unnecessary-waiting */
```
RGPosadas: We don't need this anymore as I took out the `.wait(int)` in a previous PR

RGPosadas (approved):
Tests + linting pass, LGTM!

apps/mull-ui-e2e/src/integration/US-2.7.spec.ts
cristian-aldea: will remove

PR #222, opened 2021-02-21
SONARCLOUD - Sprint 7

cindyslittleplanet:
This PR aims to fix all issues brought up by SonarCloud. 

You can access the board [here](https://sonarcloud.io/dashboard?id=RGPosadas_Mull).

Some of the issues fix include: 
- Resolving all code smells
- Increasing the code coverage
- Excluding `user.mockdata.ts` since it kept bringing up false security hotspots and false code duplication

### DIFF
--- a/apps/mull-api/src/app/event/event.resolver.ts
+++ b/apps/mull-api/src/app/event/event.resolver.ts
@@ -86,6 +86,6 @@ export class EventResolver {
 
   @ResolveField(/* istanbul ignore next */ () => Media)
   async image(@Parent() event: Event) {
-    return await this.eventService.getEventImage(event.id);
+    return this.eventService.getEventImage(event.id);
   }
 }
--- a/apps/mull-api/src/app/user/user.resolver.ts
+++ b/apps/mull-api/src/app/user/user.resolver.ts
@@ -3,7 +3,7 @@ import { Args, Int, Mutation, Parent, Query, ResolveField, Resolver } from '@nes
 import { GraphQLUpload } from 'apollo-server-express';
 import { FileUpload } from 'graphql-upload';
 import { AuthenticatedUser, AuthGuard } from '../auth/auth.guard';
-import { User } from '../entities';
+import { Media, User } from '../entities';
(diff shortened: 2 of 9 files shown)


apps/mull-api/src/app/event/event.resolver.ts
```diff
@@ -86,6 +86,6 @@ export class EventResolver {
 
   @ResolveField(/* istanbul ignore next */ () => Media)
   async image(@Parent() event: Event) {
-    return await this.eventService.getEventImage(event.id);
+    return this.eventService.getEventImage(event.id);
```
r-saba, line 89: I wasn't aware that `async` wraps returns with a promise for us and we don't have to explicitly return a promise. Good to know!

apps/mull-api/src/app/user/user.resolver.ts
```diff
@@ -39,29 +39,38 @@ export class UserResolver {
...
-    @Args('newAvatar', { type: /* istanbul ignore next */ () => GraphQLUpload, nullable: true })
-    newAvatar: FileUpload
-  ) {
-    const user = await this.userService.getUser(userInput.id);
+  private async handleNewAvatar(
+    newAvatar: FileUpload,
+    user: User,
+    userInput: UpdateUserInput
+  ): Promise<UpdateUserInput> {
+    var media: Media;
```
r-saba:
```suggestion
    let media: Media;
```

apps/mull-api/src/app/user/user.resolver.ts
```diff
@@ -39,29 +39,38 @@ export class UserResolver {
...
     if (newAvatar) {
       if (user.avatar) {
         var oldMediaId = user.avatar.id;
```
r-saba:
```suggestion
        const oldMediaId = user.avatar.id;
```

r-saba (approved):
LGTM! Left 2 changes that should be fixed but no re-request necessary.

RGPosadas (approved):
Thanks for fixing all this!

Issue #185, opened 2021-01-30
US-3.2: Group Chat Page 


JimLau49:
**User Story Description**

As a participant, I want to have a group chat with the event members so that I can communicate with them.

**User Acceptance Flow**

1. User navigates to messaging page (announcements + group messaging) 
1. User views past messages from other participants 
1. User types a message and submits it
1. Message appears on the group messaging page 

**Acceptance Criteria**

- [ ] User is able to submit a message
- [ ] User is able to see existing messages
- [ ] Messaging is done in real-time
- [ ] Seeder is updated for new entities
- [ ] User Acceptance Flow is respected
- [ ] Documentation present where necessary
- [ ] Unit and UI tests are implemented for every component and function
- [ ] There must exist an automated Cypress acceptance test that tests the user acceptance flow
- [ ] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=1749%3A28348)

**Hints**
- https://www.apollographql.com/docs/apollo-server/data/subscriptions/

RGPosadas:
https://www.apollographql.com/docs/react/data/subscriptions/#subscribing-to-updates-for-a-query

Issue #156, opened 2021-01-17
US-3.1: Announcements Page


JimLau49:
**User Story Description**

As a host, I want the ability to make announcements so that I am able to keep participants up to date on my event.

**User Acceptance Flow**

1. Host types an announcement
2. Host clicks submit
3. Announcement appears on the announcement page

**Subtasks**
- [x] 3.1.1 Creating a new event creates an Announcements channel for that event.
- [x] 3.1.2  Text input is available only to host/cohosts (participants cannot add to the announcements chat).
- [x] 3.1.3 All previous announcements are queried/displayed and then a subscription is opened to show new announcements as they come. **(Ragith & Cindy)**
- [x] 3.1.4 Hosts can post messages (must verify that user is a host before saving message in the back-end) **(Ragith)**

**Acceptance Criteria**

- [ ] Host is able to post an announcement 
- [ ] Announcement is sent to users that are in the event 
- [ ] Participants of the event cannot send messages in announcements
- [ ] Seeder is updated for new entities
- [ ] User Acceptance Flow is respected
- [ ] Documentation present where necessary
- [ ] Unit and UI tests are implemented for every component and function
- [ ] There must exist an automated Cypress acceptance test that tests the user acceptance flow
- [ ] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=2087%3A74417)

RGPosadas:
https://www.apollographql.com/docs/react/data/subscriptions/#subscribing-to-updates-for-a-query

Issue #238, opened 2021-02-25
TASK-70: Chat Bubble List Reusable component


r-saba:
**Task Description**

As a developer, I want a component to render a list of messages. 
This component accepts a list of messages and renders a list of chat-bubbles.

**Acceptance Criteria**

- [ ] Renders a list of chat-bubbles
- [ ] chat-bubbles of current user have a different style to easily identify current user messages
- [ ] Documentation present where necessary
- [ ] Unit and UI tests are implemented for every new component and function
- [ ] The application should still work with seeded data
- [ ] Must match [Mockup](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=2087%3A74417)

**Note**: This component does not query for messages, it receives a list of message data as a prop

PR #239, opened 2021-02-26
Fix: Chat Page Inputs

r-saba:
After discussion with some of the team, we agreed to remove swipeable routes for chat pages due to a known issue with position fixed and CSS transform.

This PR does the following:
- Remove swipeable routes from messages so that we can have a `position: fixed` chat input in message pages.
- Rename `MessagePage` to `ChatPagesHeader`.
- Move chat input into `AnnouncementsPage` page and remove it from `ChatPagesHeader` . This allows each chat page to control the behaviour of the chat input instead of having to do so based on the URL.
- Add logic to create default channels when a new event is created. Needed for #156 and #185


bit172 (changes requested):
Looks Good! Just a small visual bug on mobile.
This might be related to #238 but the scroll should be on the chat list and not the page. Not a blocker. @ys-lin, make sure you implement the scrollable part.
image


The input box shifted to the right on mobile.
image

r-saba:
> Looks Good! Just a small visual bug on mobile.

Thanks, fixed the visual bug.

> This might be related to #238 but the scroll should be on the chat list and not the page. Not a blocker. @ys-lin, make sure you implement the scrollable part.

We can keep it in the page too, that way each page has finer control over it. In announcements page I want to hide it when the user is not a host for example. Yes, this doesn't block this PR but we should agree on it for #238.

bit172 (approved):
LGTM!

leojrsilao (approved):
LGTM 🚀 🚀 ! 
Nice refactoring of code in the frontend, and adjusted logic for event creations with default channels in the backend
- Newly created events now have default channels
- Unit Tests and Linters pass in the front end and backend

PR #240, opened 2021-02-26
Allow easy host change

### LINKED ISSUE #217: FIX: Allow easy change of host
Currently, when we serve our front-end and back-end applications, it is served on `localhost` by default. All of our environment variables assume that the host is localhost.

However, if you serve your application on `0.0.0.0`, or your device's internal IP, you can then access the application from any device on your local network, which makes it easy to test on different devices.

Instead of hardcoding the host environment variable, we can set it in our .env file so that we don't need to change `http://localhost` everywhere where its written.

cristian-aldea:
closes #217 

The host.env.ts file was added to allow the developer to easily change the hostname where their application is being served, to allow for use cases where they don't want to run on localhost. For example, if they want other devices on their local network to be able to access the application, using localhost won't work, and they would need to host in on their machine's IP address for example

The README was updated to include info on how this mechanism works

The reason I went with a .ts file instead of adding these variables to our .env file is that the frontend cannot read the environment variables of our web server, since it is served to the client's browser. My solution, therefore, involved creating a ts in our library, which would allow us to set these values for both apps without hardcoding them.

How to test:
1. Follow instructions in the readme
2. Set your hostname to localhost and make sure everything works as before
3. Set your hostname to something else (for example, your machine's IP address) and test if you can access the app from other devices). Don't forget the --host flag!


README.md
```diff
@@ -79,9 +79,26 @@ Some secrets and sensitive configuration files are needed to properly operate ce
...
+      "ACCESS_TOKEN_SECRET": "<secret-key>"
+   }
+   ```
+
+1. Add a `host.env.ts` under the `libs/types/src/` folder. This file specify the hostnames where the application are currently being hosted, and makes it easy to change them.
+
+   ```ts
+   export const UI_HOSTNAME = 'hostname';
+   export const API_HOSTNAME = 'hostname';
+   ```
```
osahtout: is there a particular reason why these are not added to .env file? I'm not sure there's a need to add another file.

osahtout: I also noticed that all but one uses http://. I think we can set it by default to be http://localhost: in the .env file and developers can change the localhost to what ever they want without the need to add another file.

apps/mull-ui/src/app/components/event-card/event-card.spec.tsx
```diff
@@ -27,12 +26,6 @@ describe('EventCard', () => {
     expect(baseElement).toBeTruthy();
   });
 
-  it('should match snapshot', () => {
```
osahtout: wait why was this removed?

README.md
```diff
@@ -79,9 +79,26 @@ Some secrets and sensitive configuration files are needed to properly operate ce
...
+
 1. Start the server and client dev server in another terminal to test that everything is working correctly.
    - `nx serve mull-api`
-   - `nx serve mull`
+   - `nx serve mull-ui`
+   - If you specified a host other than `localhost` in your `host.env.ts`, then make sure to add the `--host` flag so that the webserver can be configured properly to serve the app on the right host
+     - example: `nx sere mull-ui --host 123.123.123.123`
```
r-saba:
```suggestion
     - example: `nx serve mull-ui --host 123.123.123.123`
```

README.md
```diff
@@ -79,9 +79,26 @@ Some secrets and sensitive configuration files are needed to properly operate ce
...
 
+1. Add a `cypress.env.json` in the `apps/mull-ui-e2e/` project folder. See `docs-and-resources` channel on slack for our secret key.
+
+   ```properties
+   {
+      "ACCESS_TOKEN_SECRET": "<secret-key>"
+   }
+   ```
+
+1. Add a `host.env.ts` under the `libs/types/src/` folder. This file specify the hostnames where the application are currently being hosted, and makes it easy to change them.
```
r-saba:
```suggestion
1. Add a `host.env.ts` under the `libs/types/src/` folder. This file specifies the hostnames where the application are currently being hosted, and makes it easy to change them.
```

r-saba (approved):
LGTM! Works both with localhost and my computers ip address. Able to access the local website from other devices when using my ipv4 address. 

Left some documentation corrections to fix before merging.

README.md
osahtout: Oh dang, I must have just skipped that part of the description sorry. 

can I ask why do we need to ignore the ts file if we just set the two string to 'http://localhost' ?

cristian-aldea: As discussed on slack, I'll be going with your suggestion @osahtout. I'll be removed the file from gitignore so that other don't need to set it up. We'll just need to keep in kind to not commit changes to that file.

cristian-aldea: However, I think its fine to keep the protocol out of the name (http://), since its not what I want to make configurable, just  the host name

apps/mull-ui/src/app/components/event-card/event-card.spec.tsx
cristian-aldea: Will be putting this back 🚀

README.md
cristian-aldea: thanks for looking out! Sorry for the sloppy readme updates on my end 😅

RGPosadas (approved):
Works perfectly. LGTM!

osahtout (approved):

PR #250, opened 2021-03-04
TASK-71: DM's Backend

### LINKED ISSUE #247: TASK-71: DMs Backend
**Task Description**

As a developer, I want to setup the DMs backend.

**Subtasks**

- [x] 71.1 Single table inheritance for DirectMessageChannel and EventChannel
- [x] 71.2 Users are related to channels

**Acceptance Criteria**

bit172:
This PR closes #247.

**Steps to Test**

1. Open your favorite GraphQL playground
2. Make sure you have users in your database
2. Use the createDm/EventChannel mutation & getDm/EventChannel query
    - getDmChannel should return null if the users do not have a channel already
3. Check if posts behave the same as before (creating, subscribing, querying)
    - The placeholder subscriber is present on the announcement page

### DIFF
--- a/apps/mull-api/src/app/channel/channel.resolver.spec.ts
+++ b/apps/mull-api/src/app/channel/channel.resolver.spec.ts
@@ -5,7 +5,6 @@ import {
   mockAllChannels,
   mockAllDirectMessageChannels,
   mockAllEventChannels,
-  mockCreateEventChannel,
 } from './channel.mockdata';
 import { ChannelResolver } from './channel.resolver';
 import { ChannelService } from './channel.service';
@@ -77,11 +76,6 @@ describe('ChannelResolver', () => {
       mockAllUsers[2].id
     );
     expect(directMessageChannel).toBe(null);
-  });
-
-  it('should create an event channel', async () => {
-    const createdEventChannel = await resolver.createEventChannel(mockCreateEventChannel);
-    expect(createdEventChannel).toBeTruthy();
   });
 
   it('should create a direct message channel', async () => {
--- a/apps/mull-api/src/app/channel/channel.resolver.spec.ts
+++ b/apps/mull-api/src/app/channel/channel.resolver.spec.ts
@@ -4,27 +4,24 @@ import { MockType } from '../user/user.service.spec';
 import {
   mockAllChannels,
   mockAllDirectMessageChannels,
-  mockCreateDmChannel,
+  mockAllEventChannels,
   mockCreateEventChannel,
 } from './channel.mockdata';
 import { ChannelResolver } from './channel.resolver';
 import { ChannelService } from './channel.service';
+import { CreateEventChannelInput } from './inputs/channel.input';
 
 const mockChannelService = () => ({
   getChannel: jest.fn((channelId: number) => {
     return mockAllChannels.find((channel) => channel.id === channelId);
   }),
-  getChannelByEvent: jest.fn((eventId, channelName, userId) => {
-    return mockAllChannels.find(
+  getChannelByEventId: jest.fn((eventId, channelName, userId) => {
+    return mockAllEventChannels.find(
       (channel) =>
         channel.event.id === eventId &&
         channel.name === channelName &&
         channel.event.host.id === userId
     );
-  }),
-  createChannel: jest.fn((mockChannelData: CreateChannelInput) => ({ ...mockChannelData })),
-  getEventChannel: jest.fn((channelId: number) => {
-    return mockAllEventChannels.find((channel) => channel.id === channelId);
   }),
   getDmChannel: jest.fn((channelId: number) => {
     return mockAllDirectMessageChannels.find((channel) => channel.id === channelId);
@@ -32,11 +29,11 @@ const mockChannelService = () => ({
   createEventChannel: jest.fn((mockEventChannelData: CreateEventChannelInput) => ({
     ...mockEventChannelData,
   })),
-  createDmChannel: jest.fn((mockChannelData: CreateDmChannelInput) => ({ ...mockChannelData })),
+  createDmChannel: jest.fn().mockReturnValue(mockAllDirectMessageChannels[0]),
   deleteChannel: jest.fn((channelId: number) =>
     mockAllChannels.find((channel) => channel.id === channelId)
   ),
-  findUsersInDmChannel: jest.fn(),
+  findDirectMessageChannelByUserIds: jest.fn(),
 });
 
 describe('ChannelResolver', () => {
@@ -55,34 +52,30 @@ describe('ChannelResolver', () => {
     expect(resolver).toBeDefined();
   });
 
-  it('should get a generic channel', async () => {
-    const getChannel = await resolver.getChannel(mockAllChannels[0].id);
-    expect(getChannel).toBe(mockAllChannels[0]);
-  });
-
   it('should get a channel by event id', async () => {
-    const getChannel = await resolver.getEventChannel(
-      mockAllChannels[2].event.id,
+    const getChannel = await resolver.getChannelByEventId(
+      mockAllEventChannels[2].event.id,
       'Announcements',
-      mockAllChannels[2].event.host.id
+      mockAllEventChannels[2].event.host.id
     );
-    expect(getChannel).toBe(mockAllChannels[2]);
-  });
-
-  it('should create a channel', async () => {
-    const createChannel = await resolver.createChannel(mockCreateChannel);
-    expect(createChannel).toBeTruthy();
+    expect(getChannel).toBe(mockAllEventChannels[2]);
   });
 
   it('should get a direct message channel', async () => {
-    service.findUsersInDmChannel.mockReturnValue(mockAllDirectMessageChannels[0]);
-    const directMessageChannel = await resolver.getDmChannel([mockAllUsers[0], mockAllUsers[1]]);
+    service.findDirectMessageChannelByUserIds.mockReturnValue(mockAllDirectMessageChannels[0]);
+    const directMessageChannel = await resolver.getDmChannel(
+      mockAllUsers[0].id,
+      mockAllUsers[1].id
+    );
     expect(directMessageChannel).toBe(mockAllDirectMessageChannels[0]);
   });
 
   it('should not get a direct message channel', async () => {
-    service.findUsersInDmChannel.mockReturnValue(null);
-    const directMessageChannel = await resolver.getDmChannel([mockAllUsers[0], mockAllUsers[2]]);
+    service.findDirectMessageChannelByUserIds.mockReturnValue(null);
+    const directMessageChannel = await resolver.getDmChannel(
+      mockAllUsers[0].id,
+      mockAllUsers[2].id
+    );
     expect(directMessageChannel).toBe(null);
   });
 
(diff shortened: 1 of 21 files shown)


cindyslittleplanet (approved):
LTGM! 🚀 Good job! 


**Subtasks**
- [x] 71.1 Single table inheritance for DirectMessageChannel and EventChannel
- [x] 71.2 Users are related to channels

**Acceptance Criteria**

apps/mull-api/src/app/channel/inputs/channel.input.ts
```diff
@@ -8,10 +10,17 @@ export class ChannelInput implements Partial<Channel> {
...
+export class CreateEventChannelInput implements Partial<EventChannel> {
   @Field()
   name: string;
 
   @Field()
   rights: number;
 }
+
+@InputType()
+export class CreateDmChannelInput implements Partial<DirectMessageChannel> {
```
osahtout: I know this may seem like I'm bias but I would suggest either have DM everywhere so `CreateDmChannelInput` and `DmChannel` and write somewhere that DM stands for direct messages or have have the full name everywhere: `CreateDirectMessageChannelInput ` and `DirectMessageChannel`

apps/mull-api/src/app/channel/channel.service.ts
```diff
@@ -1,26 +1,59 @@
...
+  async createDmChannel(input: CreateDmChannelInput): Promise<DirectMessageChannel> {
+    return this.dmChannelRepository.save(input);
+  }
+
   async deleteChannel(channelId: number): Promise<boolean> {
     await this.channelRepository.delete(channelId);
     return true;
   }
+
+  async findUsersInDmChannel(userId1: number, userId2: number): Promise<DirectMessageChannel> {
```
osahtout: The name is a little confusing. I thought this returns the users of a dm channel but it seems like it returns the channel given the users so it would be something like findDMChannelByUsersIds.

apps/mull-api/src/app/channel/channel.service.ts
```diff
@@ -1,26 +1,59 @@
...
+      .where('participants.id = :userId1 OR participants.id = :userId2', { userId1, userId2 })
+      .getMany();
+
+    // TODO: redo this filtering as TypeORM syntax
+    return dmChannelsList.find(({ participants }) => {
+      let count = 0;
+      for (const participant of participants) {
+        if (participant.id === userId1 || participant.id === userId2) count++;
+      }
+      return count === 2;
```
osahtout, line 100: This isn't a blocker just maybe a suggestion if we have time to refactor this. 

Can we have a column that concatenates the users ids to find the channel with a separator in the middle?  
Say we have user with id 1 and user with id 2, then we add to the database the string `1+2`. say we have a user with id 111 and user with id 11 then we would have the string `11+111`. we would always have the smallest number first .

Maybe it's a bad idea I'm not sure. Just wanted to hear your opinions

leojrsilao (approved):
LGTM!! 🚀 

Linters and unit tests pass.
DM Mutations are functional and users are still related to channels as intended.

apps/mull-api/src/app/channel/channel.resolver.ts
```diff
@@ -1,20 +1,27 @@
...
 export class ChannelResolver {
   constructor(private readonly channelService: ChannelService) {}
 
   @Mutation(/* istanbul ignore next */ () => Boolean)
   @UseGuards(AuthGuard)
-  async createChannel(
-    @Args('input') input: CreateChannelInput,
+  async createEventChannel(
+    @Args('input') input: CreateEventChannelInput,
     @Args('eventId', { type: /* istanbul ignore next */ () => Int }) eventId?: number
```
r-saba: `eventId` is mandatory now, we should remove the second optional argument here and have `eventId` as part of `CreateEventChannelInput`

apps/mull-api/src/app/channel/channel.resolver.ts
```diff
@@ -1,20 +1,27 @@
...
   ) {
-    return this.channelService.createChannel(input, eventId);
+    return this.channelService.createEventChannel(input, eventId);
+  }
+
+  @Mutation(/* istanbul ignore next */ () => DirectMessageChannel)
```
r-saba, line 10: Noticed here we return a Channel but in `createEventChannel` we return a Boolean. Should we make it so both return a channel or both return a boolean to be consistent?

apps/mull-api/src/app/channel/inputs/channel.input.ts
```diff
@@ -8,10 +10,17 @@ export class ChannelInput implements Partial<Channel> {
...
+  @Field(() => [UserInput])
+  @ArrayMaxSize(2)
```
r-saba: We should also at add a Min here. I can create a DM channel with 0 or 1 participant

apps/mull-api/src/app/channel/channel.resolver.ts
```diff
@@ -28,4 +35,21 @@ export class ChannelResolver {
   async getChannel(@Args('id', { type: /* istanbul ignore next */ () => Int }) id: number) {
     return this.channelService.getChannel(id);
   }
```
r-saba: Do you think we still need this or is it just unnecessary code now? Lets continue this discussion here https://github.com/RGPosadas/Mull/pull/250#discussion_r588926946

apps/mull-api/src/app/channel/channel.service.spec.ts
```diff
@@ -1,50 +1,128 @@
...
     expect(service).toBeDefined();
   });
 
-  it('should get a channel', async () => {
-    const getChannel = await service.getChannel(mockAllChannels[0].id);
-    expect(getChannel).toBe(mockAllChannels[0]);
+  it('should get a generic channel', async () => {
+    const returnedChannel = await service.getChannel(mockAllChannels[0].id);
+    expect(returnedChannel).toBe(mockAllChannels[0]);
+  });
```
r-saba, line 76: The way I was thinking about `channel`, is that it would be abstract class and we'd have other channels that implement the interface. Feel like this would be a clean and OO approach. What do you think, do we ever really want to access a channel that doesn't implement the generic one?

apps/mull-api/src/app/channel/channel.resolver.ts
```diff
@@ -28,4 +35,21 @@ export class ChannelResolver {
...
+
+  @Query(/* istanbul ignore next */ () => DirectMessageChannel, { nullable: true })
+  @UseGuards(AuthGuard)
+  async getDmChannel(
+    @Args('participants', { type: /* istanbul ignore next */ () => [UserInput] })
+    participants: UserInput[]
+  ) {
+    const [usr1, usr2] = participants;
+    const dmChannel = await this.channelService.findUsersInDmChannel(usr1.id, usr2.id);
+    return dmChannel ? dmChannel : null;
```
r-saba: This isn't secure. Since it's using AuthGuard, any user can query an DM channel by passing userId 1 and UserId 2.
We should use `AuthenticatedUser` for userId 1

apps/mull-api/src/app/channel/channel.resolver.ts
```diff
@@ -28,4 +35,21 @@ export class ChannelResolver {
   async getChannel(@Args('id', { type: /* istanbul ignore next */ () => Int }) id: number) {
     return this.channelService.getChannel(id);
   }
+
+  @Query(/* istanbul ignore next */ () => EventChannel)
+  @UseGuards(AuthGuard)
+  async getEventChannel(@Args('id', { type: /* istanbul ignore next */ () => Int }) id: number) {
+    return this.channelService.getEventChannel(id);
+  }
```
r-saba: How do we want to handle security for event channels? In #251 I implemented the logic already so it would be redundant for you to do it again here

r-saba (changes requested):
Blocking because we should handle security of mutations/queries in this PR too

apps/mull-api/src/app/channel/channel.resolver.ts
r-saba: We want to make sure security is handled for all DM channel queries/mutations

apps/mull-api/src/app/channel/channel.resolver.ts
bit172: Sure, we will return it for all creates/deletes in case we need it in the future

apps/mull-api/src/app/channel/channel.service.spec.ts
bit172: We will delete this resolver.

apps/mull-api/src/app/channel/channel.resolver.ts
bit172: Thank you for bringing this up, we will check for security in our resolvers

apps/mull-api/src/app/channel/channel.resolver.ts
RGPosadas: Done!

apps/mull-api/src/app/channel/channel.service.ts
RGPosadas: After talking, we decided not to do this.

apps/mull-api/src/app/channel/channel.service.ts
RGPosadas: Done

apps/mull-api/src/app/channel/channel.resolver.ts
RGPosadas: Done

RGPosadas: Upon further inspection, createEventChannel does not need to exposed in their respective service/resolver files as createEvent is the one creating those channels through `cascade: true`. Therefore, we will be removing these functions as they are not used anywhere else.

leojrsilao (approved):

apps/mull-api/src/app/channel/channel.resolver.spec.ts
```diff
@@ -4,39 +4,36 @@ import { MockType } from '../user/user.service.spec';
...
-    return mockAllEventChannels.find((channel) => channel.id === channelId);
-  }),
   getDmChannel: jest.fn((channelId: number) => {
     return mockAllDirectMessageChannels.find((channel) => channel.id === channelId);
   }),
   createEventChannel: jest.fn((mockEventChannelData: CreateEventChannelInput) => ({
     ...mockEventChannelData,
   })),
-  createDmChannel: jest.fn((mockChannelData: CreateDmChannelInput) => ({ ...mockChannelData })),
+  createDmChannel: jest.fn().mockReturnValue(mockAllDirectMessageChannels[0]),
```
r-saba: Can we implement the logic instead of hard coding the return value? We've done this in other tests and it's made the debugging process/reusability of these mock implementations difficult. If it doesn't take too much time.

r-saba (changes requested):
Noticed I can create a multiple DM channels with the same list of participants. This would be an issue when trying to retrieve a DM channel between two users.

apps/mull-api/src/app/channel/channel.service.spec.ts
```diff
@@ -1,66 +1,148 @@
...
+
+  it('should throw an exception when creating a DM channel with 2 users that already have a DM channel', async () => {
+    directMessageRepository.createQueryBuilder.mockReturnValue({
+      where: jest.fn().mockReturnThis(),
+      leftJoinAndSelect: jest.fn().mockReturnThis(),
+      getMany: jest.fn().mockReturnValue(mockAllDirectMessageChannels),
+    });
+    expect(
+      async () => await service.createDirectMessageChannel(mockAllUsers[0].id, mockAllUsers[1].id)
+    ).rejects.toThrow(UnauthorizedException);
```
r-saba: Some error messages resulted from this test. To fix before merging
```suggestion
    ).rejects.toThrow(ConflictException);
```

r-saba (approved):
LGTM, left a small comment to fix before merging
**Acceptance Criteria**
- [X] Documentation present where necessary
- [X] Unit and UI tests are implemented for every new component and function
- [X] The application should still work with seeded data

PR #261, opened 2021-03-11
US-1.6: Event Restriction

### LINKED ISSUE #255: US-1.6: Event Restriction
**User Story Description**

As a host, I want to control whether my event is public, friends only, or private, so that I can manage who can see and join my event.

**User Acceptance Flow**

1. User clicks on `Create Event`
1. User is able to choose the event restriction (Public, Friends Only, Private)
1. User creates event
    1. Only public events can be discovered on Discover Events Page 
    1. Non-public events cannot be seen on Discover Events page

**Acceptance Criteria**

- [ ] User Acceptance Flow is respected
- [ ] There must exist an automated Cypress acceptance test which tests the user acceptance flow

r-saba:
Closes #255

In these e2e tests, we test that public events are visible on the discover page and that private events are not.

**Notes**
I ran into an issue when using `mockRefreshRequest` to create event, the event would not be visible on the discover tab in the e2e even though the test would pass. I felt that not having the visual feedback of seeing the event on the discover tab in our gifs would be unsatisfactory. I decided to register and login before the tests instead of using `mockRefreshRequests`. The tests still pass and we can see the event on the discover tab which would be useful for the gif

### DIFF
--- a/apps/mull-ui-e2e/src/integration/US-1.6.spec.ts
+++ b/apps/mull-ui-e2e/src/integration/US-1.6.spec.ts
@@ -1,0 +1,107 @@
+import 'cypress-file-upload';
+import * as faker from 'faker';
+import { geolocationStub } from '../fixtures';
+import { frameSizes } from './../fixtures/frame-sizes';
+
+/**
+ * Helper function to register
+ * @param email
+ * @param password
+ */
+const registerUser = (email: string, password: string) => {
+  cy.visit('http://localhost:4200/register');
+  cy.get('#name').type('Tyler Uno');
+  cy.get('#email').type(email);
+  cy.get('#password').type(password);
+
+  cy.get('.register-button').click();
+  cy.get('.Toastify__toast.Toastify__toast--success', { timeout: 5000 }).should(
+    'have.css',
+    'background-color',
+    'rgb(39, 176, 154)'
+  );
+};
+/**
+ * Helper function to login
+ * @param username
+ * @param password
+ */
+const loginUser = (username: string, password: string) => {
+  cy.get('#email').type(username);
+  cy.get('#password').type(password);
+  cy.get('.login').click();
+  cy.get('.Toastify__toast.Toastify__toast--success', { timeout: 5000 }).should(
+    'have.css',
+    'background-color',
+    'rgb(39, 176, 154)'
+  );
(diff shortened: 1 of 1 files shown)


JimLau49 (approved):
LGTM! Thanks for also providing the documentation as well :) 

**Acceptance Criteria**

- [x] User Acceptance Flow is respected
- [x] There must exist an automated Cypress acceptance test which tests the user acceptance flow


image

apps/mull-ui-e2e/src/integration/US-1.6.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+  cy.get('#eventTitle').type(date.toString());
+
+  cy.get('.-today').click();
+  cy.get('.-today').click();
+
+  cy.get('#description').type('test description');
+
+  cy.get('#location').click();
+  cy.get('#location-input-field').should('be.visible');
+
```
RGPosadas, line 60: The tests were really flaky for me when using Current Location. Each test at some point passed, but I never had a full run pass completely.

I looked into why US-1.1 always passes for me, and it was because we put in a location as such:
```
cy.get('#location-input-field').type('845 Rue Sherbrooke');
      cy.get('#location-input-field-option-0', { timeout: 3500 }).should(
        'contain.text',
        '845 Rue Sherbrooke'
      );
```

Adding this ensured this E2E test was no longer flaky for me.

RGPosadas (approved):
Other than the flakiness, everything looks good! However, make sure to fix the linting errors.

**Acceptance Criteria**
- [x] User Acceptance Flow is respected
- [x] There must exist an automated Cypress acceptance test which tests the user acceptance flow

apps/mull-ui-e2e/src/integration/US-1.6.spec.ts
r-saba: I decided not to do the above to not make unnecessary API calls. It isn't flaky for me, all tests consistently pass. 
I added a geolocation stub which might help with your issue. Let me know if it works, if not I'll add the API call.

RGPosadas: This resolved my issue. Thanks!

PR #287, opened 2021-03-14
Add troubleshoot info for non localhost dev build

cristian-aldea:


cristian-aldea:
~~Will also add instructions to get ssl and https working for localhost, since that might be needed for some features~~

~~The `nx serve` command supports SSL: https://nx.dev/latest/react/cli/serve~~

Nevermind, was way too complicated to setup. Not worth it

JimLau49 (approved):
followed the instructions and was able to host on my ip address

osahtout (approved):

Issue #273, opened 2021-03-13
US-8.1: Add Friends


RGPosadas:
**User Story Description**

As a user, I want the ability to add friends by searching for them so that I may participate in events with them

**User Acceptance Flow**

1. User logs in
1. User goes to their Profile Page
1. User clicks on 'Add Friends'
1. Add Friends page shows up where they can search for friends, accept friend requests, and see pending friend requests

**Acceptance Criteria**

- [ ] User Acceptance Flow is respected
- [ ] There must exist an automated Cypress acceptance test which tests the user acceptance flow
- [ ] Documentation present where necessary
- [ ] Unit and UI tests are implemented for every new component and function
- [ ] The application should still work with seeded data
- [ ] Must match mockups:
   - "[Added Me + Pending Requests](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=1596%3A5062)" view
   - "[Pending Request modal](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=1695%3A83663)" view
   - "[Search Users](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=1695%3A86144)" view

PR #302, opened 2021-03-27
TASK-79: Add friends backend

### LINKED ISSUE #277: TASK-79: Add Friends Backend
**Task Description**

As a developer, I want to create the backend queries to support the pages for the frontend (#276) .

**Acceptance Criteria**

- [ ] Queries exist:
   - addFriend
   - removeFriend
   - getRelationships
- [ ] Security is handled in the backend

cristian-aldea:
closes #277

This PR aims to implement a few backend functions to be used in the features of #273.
   - addFriend
   - removeFriend
   - getRelationships

How to test:
- Make sure you have at least 2 users in your DB.
- Run the backend
- Try to use the addFriend and removeFriend mutations and check what happens when you call the `friends` query. It should behave as expected. addFriend should be a one-sided add, equivalent to sending a request. removeFriend should be a two-way remove.
- Call the getRelationships query. This returns a list of Relationships (see added class). What I recommend is going into MySQL and modifying the rows manually and making sure that the list returned by the query changes as expected.

Examples (all from perspective of user id 1):
No relationships:
image

user 1 added user 2
image

user 2 added user 1
image

Both users added each other
image

Works with multiple users too
image


apps/mull-api/src/app/user/user.service.ts
```diff
@@ -73,4 +75,87 @@ export class UserService {
     await this.userRepository.save(currentUser);
     return true;
   }
+
+  async removeFriend(currentUserId: number, userIdToRemove: number): Promise<boolean> {
+    if (currentUserId === userIdToRemove) {
+      throw new UnprocessableEntityException('Cannot remove oneself as friend.');
+    }
```
r-saba, line 82: Should we also check that `userIdToRemove` is an actual user. Noticed we're not checking in the `addFriend` function as well which is a more serious issue as I believe it could break the UI.

apps/mull-api/src/app/utilities.ts
```diff
@@ -0,0 +1,11 @@
+export const setUnion = <T>(a: Set<T>, b: Set<T>) => {
```
r-saba, line 1: Nice

apps/mull-api/src/app/user/user.service.ts
cristian-aldea: I think its not necessary, it would add an extra query (I think), and I don't see why we should send back an error if we want to remove someone from your friends who never was your friend to begin with (because they're a ghost)

leojrsilao (approved):
LGTM!! 🚀 🚀 
Mutations and queries for adding friends are functional.
Unit tests and linters pass in the backend.

apps/mull-api/src/app/user/user.service.ts
r-saba: Yes, I think it's overkill as well. Lets keep it as is.

r-saba (approved):
LGTM! All queries, mutations,test and linter worked as expected.

PR #308, opened 2021-03-30
Fix problems in camera feature implementation

### LINKED ISSUE #290: FIX: Better iOS support for ml features
**Describe The Bug**

Currently, our ML features present a few issues on iOS which need to be resolved if the feature is to be used by iOS users.
1. the modal doesn't seem to be working. 
2. iOS (and maybe other browsers/platforms) don't allow autoplay. [Link](https://stackoverflow.com/questions/31776548/why-cant-javascript-play-audio-files-on-iphone-safari)
3. There is weird interaction with the video element and the rest of the page. The video element seems to go fullscreen and therefore the whole feature doesn't work since the canvas is now behind the video.

Solutions:
1. Check if backend is available. If not, give an appropriate error message to the user
2. Add a button that must be pressed to start detection
3. Remove video element and instead just use a canvas. This might also fix issue 2.

cristian-aldea:
The camera detection feature suffers from a few issues on iOS which made it impossible to work properly on that platform. I'm happy to say that most of the issues have been fixed, and consequently, the overall implementation of the feature is much improved.
- Load all SVG images upfront once, which speeds up drawing and fixes many issues with icons not appearing
- Change how width/height of video stream is fetched since the previous method is newer and not well supported on some browsers
- Set video to be muted. This solves an issue on iOS where the video wouldn't start playing since iOS prohibits video autoplay under certain conditions: https://developer.apple.com/documentation/webkit/delivering_video_content_for_safari
- Set the video to be inline. This fixes an issue where the video would go fullscreen on iOS
- Rounding class output of the model. All the numbers the model returns are floats, and floating-point precision issues starting cropping out on iOS, since the class of a detection is meant to be an integer. Now, these issues will never occur again :)

How to test:
- Log in and go to the camera feature
- Try pointing the camera at supported objects (see [this variable](https://github.com/RGPosadas/Mull/blob/develop/libs/types/src/ml.ts#L24)). You should see an icon pop up
- Try pressing on the icons that show up. You should see a modal pop up.

image
image

### DIFF
--- a/apps/mull-ui/src/app/pages/waste-recognition-page/waste-recognition-page.tsx
+++ b/apps/mull-ui/src/app/pages/waste-recognition-page/waste-recognition-page.tsx
@@ -1,7 +1,7 @@
 import { DetectionResult, wasteClassMap } from '@mull/types';
 import React, { useEffect, useRef, useState } from 'react';
 import { toast } from 'react-toastify';
-import { MULL_MODEL_URL, WasteTypeSvgMap } from '../../../constants';
+import { MULL_MODEL_URL, WasteIconUrlMap } from '../../../constants';
 import { canvasToImageCoords, coordsInBox, drawDetectionIcons } from '../../../utilities';
 import { MullModal } from '../../components';
 import { useToast } from '../../hooks/useToast';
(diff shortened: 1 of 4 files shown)


JimLau49 (approved):
Tested various objects , platform: iOS 

Working flawlessly, nice work Cristian 🚀

apps/mull-ui/src/app/pages/waste-recognition-page/waste-recognition-page.tsx
```diff
@@ -80,18 +80,19 @@ export function WasteRecognitionPage(props: WasteRecognitionPageProps) {
...
-          videoRef.current.width = width;
-          canvasRef.current.height = height;
-          canvasRef.current.width = width;
-
           videoRef.current.play();
 
           return new Promise<MediaStream>((resolve, reject) => {
             videoRef.current.onloadedmetadata = () => {
+              // Doesn't work on all browsers. Keeping it here just in case
+              // const { width, height } = stream.getTracks()[0].getSettings();
```
osahtout, line 88: I would simply remove it

osahtout (approved):

apps/mull-ui/src/app/pages/waste-recognition-page/waste-recognition-page.tsx
cristian-aldea: I'd like to keep it just in case it might be relevant again.

Issue #284, opened 2021-03-13
TASK-83: Viewing Other User Profile Page Flows


RGPosadas:
**Task Description**

As a developer, I want to setup the flows that will lead to viewing another user's profile page.

Flow #1: UserA is friends with UserB
1. UserA logs in
1. UserA goes to their Profile Page
1. UserA clicks on My Friends
1. UserA clicks on UserB's profile
1. UserA sees UserB's profile page 

Flow #2: UserA is not friends with UserB
1. UserA logs in
1. UserA goes to their Profile Page
1. UserA clicks on Add Friends
1. UserA types in UserB's name
1. UserA clicks on UserB's profile
1. UserA sees UserB's profile page 

Flow #3: Accessing UserB's profile through chatting pages
1. UserA is in a chat with UserB (either Direct Messages or Event Messages)
1. UserA clicks on UserB's avatar
1. UserA sees UserB's profile

**Acceptance Criteria**
- [ ] Connect pages to each other which will eventually lead to a user's profile page
- [ ] Flows are respected
- [ ] Update E2E test as needed
- [ ] Documentation present where necessary
- [ ] Unit and UI tests are implemented for every new component and function

RGPosadas:
Closing since these flows have already been done in their respective tickets.

Issue #270, opened 2021-03-13
US-7.4: View Other User Profiles


RGPosadas:
**User Story Description**

As a user, I want to see other users’ profiles so that I may see their profile and events that they’ve participated in.

**User Acceptance Flow**
1. UserA logs in
1. UserA sees UserB's profile page

**Note**: You can hardcode it so that it will check the relationship between user 1 (current user) and user 2.

**Acceptance Criteria**

- [ ] User Acceptance Flow is respected
- [ ] There must exist an automated Cypress acceptance test which tests the user acceptance flow
- [ ] Documentation present where necessary
- [ ] Unit and UI tests are implemented for every new component and function
- [ ] The application should still work with seeded data
- [ ] Must match mockups:
   - [Add Friend ](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=1793%3A58846)view
   - [Pending](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=1793%3A58851) view
   - [Friends](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=1793%3A58834) view
   - [Remove Friend modal](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=1793%3A58839) view

PR #330, opened 2021-04-02
TASK-76: View Other User Profiles Frontend

### LINKED ISSUE #272: TASK-76: View Other User Profiles Frontend
**Task Description**

As a developer, I want to setup the frontend for viewing other users' profiles

**Acceptance Criteria**

- [ ] user-relationship buttons exist
- [ ] Modal exists
- [ ] If queries already exist, use them already instead of using placeholders (for example, getting portfolio events)
- [ ] Must match mockups:
   - [Add Friend ](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=1793%3A58846)view
   - [Pending](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=1793%3A58851) view
   - [Friends](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=1793%3A58834) view
   - [Remove Friend modal](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=1793%3A58839) view

RGPosadas:
Closes #272 

To test:
* Head on over to http://localhost:4200/other-user-profile
* Change [this line's ID](https://github.com/RGPosadas/Mull/blob/TASK-76/view-other-user-profiles-frontend/apps/mull-ui/src/app/pages/profile/other-user-profile/other-user-profile.tsx#L20) to get the user you want
   * Note that if this ID is the current authenticated user ID, then it will redirect to `/profile`. This is for when a user clicks on their avatar in a messages page, for example. The goal is for /user/user.id to redirect to this page. The whole workflow is to be implemented in #284.
* Ensure that the relationship  between the current user and the other user is accurate with respect to the `Friends` table in the db


apps/mull-api/src/app/user/user.resolver.ts
```diff
@@ -29,8 +29,9 @@ export class UserResolver {
     return this.userService.getAllUsers();
   }
 
+  @UseGuards(AuthGuard)
```
cristian-aldea, line 32: Thanks for making these changes. At first it made sense to use @authticateduser, but since these queries are needed for other users too its best to simply have the id as a variable :/

apps/mull-api/src/app/user/user.resolver.ts
```diff
@@ -96,40 +97,51 @@ export class UserResolver {
...
   }
 
+  @Mutation(/* istanbul ignore next */ () => Boolean)
+  async removePendingRequest(
+    @AuthenticatedUser() currentUserId: number,
+    @Args('userIdToRemove', { type: /* istanbul ignore next */ () => Int }) userIdToRemove: number
+  ) {
+    return this.userService.removeFriend(currentUserId, userIdToRemove);
+  }
+
```
cristian-aldea, line 144: 💡I'd say this isn't needed, you can use the `removeFriend` mutation wherever you're using `removePendingRequest`, it'll behave in the same way

apps/mull-api/src/app/user/user.resolver.ts
```diff
@@ -138,7 +150,7 @@ export class UserResolver {
   @Query(/* istanbul ignore next */ () => RelationshipType)
   async getUserRelationship(
     @AuthenticatedUser() userIdA: number,
-    @Args('userIdB') userIdB: number
+    @Args('userIdB', { type: /* istanbul ignore next */ () => Int }) userIdB: number
```
cristian-aldea, line 153: Thanks, forgot to add this haha ♥

apps/mull-api/uploads/.gitignore
```diff
@@ -1,3 +1,4 @@
 *.jpg
 *.jpeg
 *.png
+*.gif
```
cristian-aldea, line 4: Could you also add `*.webp` while you're at it 👀

apps/mull-ui/src/app/pages/profile/other-user-profile/other-user-profile.tsx
```diff
@@ -0,0 +1,59 @@
...
+export interface OtherUserProfilePageProps {
+  history: History;
+  prevPage: string;
+}
+
+export const OtherUserProfilePage = ({ history, prevPage }: OtherUserProfilePageProps) => {
+  const { data: otherUserProfileData, loading: userProfileLoading } = useOtherUserProfileQuery({
+    variables: {
+      // TODO in TASK-83: get id of other user
+      id: 4,
```
cristian-aldea, line 20: ⚠I understand you might want to address this in TASK-83, but I feel like it'd be good to do this properly now instead of reworking it later.

Instead of changing the number here in the code. You should set `OtherUserProfilePage` to show up on a parameterized route like `/profile/:id`. And simply fetch the id from the route for your query.

This would be much simpler if the profile page was one component 👀 , but since we're not doing that, we could instead have a wrapper component that shows up when we go to `/profile/:id`. It would then check if :id is the same as the current user id. if it is, then it would show the user profile, otherwise, it would show the otheruserprofile.

If you really think this shouldn't be addressed now, I'm fine with dealing with this later, but I think this will make things simpler in the end.

cristian-aldea (changes requested):
Thanks for the hard work on this. Looks great :).

I have some 💡 suggestions and one ⚠problem with how routes are done for profiles. Please come ask me if you have questions

apps/mull-api/src/app/user/user.resolver.ts
RGPosadas: I still preferred a separate mutation just for the semantic name, viewing it from the frontend

apps/mull-ui/src/app/pages/profile/other-user-profile/other-user-profile.tsx
RGPosadas: Ah I understand your concern, and you are right - it should be implemented before #284 as it will make the E2E easier to create (and less refactoring later on) as well. However, I do want to implement your suggestion in #270 if you don't mind, as there isn't much to do in that ticket anyway other than the E2E currently, and this ticket took longer than expected

apps/mull-api/uploads/.gitignore
RGPosadas: Done!

apps/mull-ui/src/app/pages/profile/other-user-profile/other-user-profile.tsx
cristian-aldea: sure, that's fine, I'll assign myself to that ticket to make sure that gets addressed then

cristian-aldea (approved):
As we discussed, my concerns will be addressed in the user story
- [x] user-relationship buttons exist
- [x] Modal exists
- [x] If queries already exist, use them already instead of using placeholders (for example, getting portfolio events)
- [x] Must match mockups:
   - [Add Friend ](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=1793%3A58846)view
   - [Pending](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=1793%3A58851) view
   - [Friends](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=1793%3A58834) view
   - [Remove Friend modal](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=1793%3A58839) view

ys-lin (approved):
- [x] user-relationship buttons exist
- [x] Modal exists
- [x] If queries already exist, use them already instead of using placeholders (for example, getting portfolio events)
- [x] Must match mockups:
   - [Add Friend ](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=1793%3A58846)view
   - [Pending](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=1793%3A58851) view
   - [Friends](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=1793%3A58834) view
   - [Remove Friend modal](https://www.figma.com/file/FJPXMYI4ULCnuCEZvPs67k/Mull?node-id=1793%3A58839) view

leojrsilao (approved):
LGTM!! 🚀 🚀 
Linting and unit tests passes.
Functionality of adding / cancelling request of a friend works on the view profile page as well as heading to their events.

PR #340, opened 2021-04-06
TASK-91: Profile Page Styling/UX Cleanup

### LINKED ISSUE #319: TASK-91: Profile Page Styling/UX Cleanup
**Task Description**

I want to fix the styling and improve the UX of the Profile page:

- [ ] 91.1  Restrict number of lines that can be added in the description despite character length

**Acceptance Criteria**

cindyslittleplanet:
This PR closes #319.

In order to ensure that a lengthy user description does not hide the profile page content, two changes were made to it:

1. A check to make sure the description does not exceed 6 lines and/or 1024 characters before saving the description.
2. If a user inputs less than 6 lines and less than 1024 characters, but the description is above 150 characters (default arbitrary number), then the description becomes an expandable text. 

I also changed the dead buttons `Portfolio`, `Friends`, `Hosting` in the profile header to plain texts since they weren't redirecting to anywhere.


apps/mull-ui/src/utilities.ts
```diff
@@ -83,6 +84,16 @@ export const hasEmoji = (text: string) => {
   return !!emojiRegexRGI().exec(text);
 };
 
+const textLength = string.length;
+export const exceedsNbOfLines = (text: string) => {
+  let countLines = 1;
+  for (let i = 0; i < textLength; i++) if (text.charAt(i) === '\n') countLines++;
+  if (countLines > 6) {
```
JimLau49: Why is this set to 6? and what exactly do lines mean as i put in 6 rows of @ and it did not reject me as it wasnt at the char limit 

image

JimLau49 (approved):
LGTM, left a question

**Acceptance Criteria:**
- [x] 91.1  Restrict number of lines that can be added in the description despite character length

image

apps/mull-ui/src/utilities.ts
cindyslittleplanet: 6 is the default number we came up with that seemed reasonable both on mobile and desktop. 
And 6 lines means 6 keyboard `enter`, not rows. Do you have any suggestion on how to call keyboard enters?

apps/mull-ui/src/app/components/profile-header/profile-header.tsx
```diff
@@ -146,30 +147,41 @@ export const ProfileHeader = ({
...
           className={
-            isCurrentUser ? 'profile-side-container for-current-user' : 'profile-side-container'
+            isCurrentUser
+              ? 'profile-side-container for-current-user'
+              : 'profile-side-container for-other-user'
           }
         >
-          <button className="profile-stats" data-testid="portfolioCount">
+          <p
+            className={isCurrentUser ? 'profile-stats' : 'profile-stats for-other-user'}
```
r-saba: I find this is less repetition and easier read
```suggestion
            className={`profile-stats ${isCurrentUser ? 'for-other-user': ''}`}
```

r-saba: Since a couple of these have the same class name, you could even extract the classname into a variable

apps/mull-ui/src/utilities.ts
```diff
@@ -83,6 +84,16 @@ export const hasEmoji = (text: string) => {
   return !!emojiRegexRGI().exec(text);
 };
 
+const textLength = string.length;
+export const exceedsNbOfLines = (text: string) => {
+  let countLines = 1;
+  for (let i = 0; i < textLength; i++) if (text.charAt(i) === '\n') countLines++;
```
r-saba: Not used to seeing for loops like this, I prefer the usual way even if it's just one line inside. My brain had to pause for a second lool
```suggestion
   for (let i = 0; i < textLength; i++) {
    if (text.charAt(i) === '\n') countLines++;
  }
```

apps/mull-ui/src/app/pages/profile/edit-profile/edit-profile.tsx
```diff
@@ -41,11 +41,19 @@ const EditProfile = ({ history }: EditProfilePageProps) => {
...
+          LIMITS.PROFILE_DESCRIPTION,
+          `Description must be ${LIMITS.PROFILE_DESCRIPTION} characters or less.`
+        )
+        .test(
+          'NbOfLinesCheck',
+          `Description must not exceed ${LIMITS.LINES} lines.`,
+          function (description) {
+            return exceedsNbOfLines(description);
+          }
+        ),
```
r-saba: es6 syntax to make it shorter and easier to read.
```suggestion
        .test(
          'NbOfLinesCheck',
          `Description must not exceed ${LIMITS.LINES} lines.`, (description) => exceedsNbOfLines(description)
        ),
```

r-saba (changes requested):
I can still add more than 6 lines, could you take a look into this and why?
image

apps/mull-ui/src/app/pages/profile/edit-profile/edit-profile.tsx
```diff
@@ -41,11 +41,19 @@ const EditProfile = ({ history }: EditProfilePageProps) => {
...
       imageFile: Yup.mixed().test('big-file', 'File size is too large', validateFileSize),
+      description: Yup.string()
+        .max(
+          LIMITS.PROFILE_DESCRIPTION,
+          `Description must be ${LIMITS.PROFILE_DESCRIPTION} characters or less.`
+        )
+        .test(
+          'NbOfLinesCheck',
+          `Description must not exceed ${LIMITS.LINES} lines.`,
+          function (description) {
```
cristian-aldea: 💡
🏹 > `function`

apps/mull-ui/src/utilities.ts
```diff
@@ -96,6 +97,16 @@ export const hasEmoji = (text: string) => {
...
+  if (countLines > 6) {
+    return false;
+  }
+  return true;
+};
```
cristian-aldea: ⚠So, a few issues

Your code here doesn't prevent me from writing a desc with more than 6 lines.
image

This is because of line 87, which you took `string.length` as the max number to loop through. This value will always be 0, since you're taking the length of the string type.

I'd recommend a much simpler approach that doesn't need a separate function or looping, using regex. Here's an example
image

cristian-aldea (changes requested):
Got some 💡 suggestions, and found an ⚠ issue I'd like addressed.

For the profile header, I'd love if we could redo the styling since its very weird, but it work for now.

cristian-aldea:
> I can still add more than 6 lines, could you take a look into this and why?

My comment [here](https://github.com/RGPosadas/Mull/pull/340#discussion_r608799471) explains why

apps/mull-ui/src/utilities.ts
cindyslittleplanet: I hastily opened the PR without double checking the functionality after adding a new line which broke the 6 lines restriction. That's my bad.
Thank you for the regex suggestion, but I tried several combination of regex and it unfortunately counts every space instead and not each new line (not to mention how the `match` added an extra console error to fix), so I decided to stick with my for loop instead.

apps/mull-ui/src/utilities.ts
```diff
@@ -96,6 +96,20 @@ export const hasEmoji = (text: string) => {
...
+    let countLines = 1;
+    for (let i = 0; i < descriptionLines.length; i++) {
+      if (descriptionLines.charAt(i) === '\n') countLines++;
+    }
+    if (countLines > 6) {
+      return false;
+    }
+    return true;
+  }
+  return true;
```
cristian-aldea: 💡This doesn't need to be this complicated:
1. Instead of checking if a variable has the type `"undefined"`, its better to check if the variable is undefined by doing a check to see if the variable is false, or equal to `undefined`
2. You can replace all of this with a regex call:

```suggestion
  // True if string falsy or < 6 newlines
  return !string || string.match(/\n/gi)?.length < 6;
```

cristian-aldea: image

cristian-aldea: and like I mentioned something this simple might not need its own function

apps/mull-ui/src/utilities.ts
cristian-aldea: > but I tried several combinations of regex and it, unfortunately, counts every space instead and not each new line

Depends on what regex you use. Did you try mine? It works pretty much exactly like your code does right now
image
image

If you did `\s` instead of `\n` then yeah it would count the number of spaces

> not to mention how the `match` added an extra console error to fix

Maybe, but that's not a big deal, have a look at my suggestion below :)

libs/types/src/forms.ts
```diff
@@ -38,4 +38,6 @@ export const LIMITS = {
   USERNAME: 64,
   EVENT_TITLE: 64,
   IMAGE_SIZE: 10000000,
+  PROFILE_DESCRIPTION: 1024,
+  LINES: 6,
```
cristian-aldea: 💡 Just so that it's clear that this is for the profile description
```suggestion
  PROFILE_DESCRIPTION_LINES: 6,
```

cristian-aldea (approved):
Everything looks great! Thanks for addressing my comments.

Left another small suggestion.

r-saba (changes requested):
Works well, can you make the padding the same all around the textarea. It looks a bit off since the right side has so much padding. I'd say make it 1em all around, only for this page though.
image

apps/mull-ui/src/app/components/expandable-text/expandable-text.tsx
```diff
@@ -1,25 +1,34 @@
...
+  className = '',
+}: ExpandableTextProps) => {
   const [expand, setExpand] = useState(false);
 
   const getText = () => {
     if (expand) {
       return children;
     } else {
-      return children.substr(0, cutoff);
+      const mincuttoff = min([cutoff, getIndexOfNthOccurence(children, '\n', maxlines)]);
```
r-saba, line 25: Good stuff!

r-saba (approved):
LGTM! Good work!

PR #342, opened 2021-04-06
FIX: Adjust image styling

### LINKED ISSUE #341: FIX: Adjust EventBullet styling
As people have noticed before, you sometimes see some oddity on the right
image

This is caused by the lack of padding on the event message row. The image is also sometimes distorted.
image

cristian-aldea:
closes #341
We weren't taking into account the use of wide and long images for the chat bubble, the file upload and the event bullet

Before:
image
image
image

After: 
image
image
image

### DIFF
--- a/apps/mull-ui/src/app/components/chat-bubble-list/chat-bubble-list.scss
+++ b/apps/mull-ui/src/app/components/chat-bubble-list/chat-bubble-list.scss
@@ -1,4 +1,6 @@
 .chat-bubble-image {
   max-width: 11rem;
+  max-height: 12rem;
   display: block;
+  object-fit: contain;
 }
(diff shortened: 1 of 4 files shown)


JimLau49 (approved):
LGTM Mr.tall isnt that tall anymore!

RGPosadas (commented):
I'm not too sure about these rows varying due to the image width. Should we make everything square and introduce padding for those that don't fit?  I am not too certain what would be best tbh
image

cristian-aldea:
> I'm not too sure about these rows varying due to the image width. Should we make everything square and introduce padding for those that don't fit? I am not too certain what would be best tbh

Honestly yeah that sounds better. I just need to add a fixed width instead of a max-width to fix that. The object-fit: contain will make sure the image isn't stretched or anything

cristian-aldea:
Thanks for the suggestion @RGPosadas, how does it look?
image

RGPosadas (approved):
Perfect! LGTM 🚀 
image

PR #334, opened 2021-04-05
Task-88: Create Event page UX Styling Cleanup

ys-lin:
Closes #316 

How to test: please check if every acceptance criteria is satisfied.

PR #344, opened 2021-04-06
FIX: Remove SwipeableRoutes in favor of Switch

### LINKED ISSUE #345: FIX: Remove swipeableroutes in favor of switch
Currently, on develop, if you go to any `/home/anything` route by either typing that in the URL box or history.push, you will be redirected to the discover page. This happens event on valid routes such as `/home/upcoming` and `/home/myevents`. This causes several issues:
- If you go to the myevents page, press on an event and then go back, you are faced with the discover page, which makes no sense and messes with the user flow
- If you need to history.push to either the upcoming or myevent pages in the code (like with [Yunshi's PR](#334), this *WILL NOT* work, since you will get redirected to the discover page
- As a user, you can't go directly to the upcoming and myevents pages. You have to start off at the discover page and then press/swipe your way to the other pages.

This is pretty bad. Is it caused by the SwipeableRoutes component. No matter not I adjust the 4th catch-all route which redirects to the discover page, the page won't work correctly. From reading their documentation, there isn't even an example with a catch-all base route: https://www.npmjs.com/package/react-swipeable-routes

```tsx
              <SwipeableRoutes>
                <PrivateRoute path={ROUTES.DISCOVER.url} component={DiscoverPage} />
                <PrivateRoute path={ROUTES.UPCOMING.url} component={UpcomingPage} />
                <PrivateRoute path={ROUTES.MY_EVENTS.url} component={MyEventsPage} />
                <PrivateRoute exact path={ROUTES.HOME}>
                  <Redirect to={ROUTES.DISCOVER.url} />
                </PrivateRoute>
              </SwipeableRoutes>
```

cristian-aldea:
closes #345 

Currently, on develop, if you go to any `/home/anything` route by either typing that in the URL box or history.push, you will be redirected to the discover page. This happens event on valid routes such as `/home/upcoming` and `/home/myevents`. This causes several issues:
- If you go to the myevents page, press on an event and then go back, you are faced with the discover page, which makes no sense and messes with the user flow
- If you need to history.push to either the upcoming or myevent pages in the code (like with [Yunshi's PR](#334), this *WILL NOT* work, since you will get redirected to the discover page
- As a user, you can't go directly to the upcoming and myevents pages. You have to start off at the discover page and then press/swipe your way to the other pages.

This is pretty bad. Is it caused by the SwipeableRoutes component. No matter not I adjust the 4th catch-all route which redirects to the discover page, the page won't work correctly. From reading their documentation, there isn't even an example with a catch-all base route: https://www.npmjs.com/package/react-swipeable-routes

```tsx
              <SwipeableRoutes>
                <PrivateRoute path={ROUTES.DISCOVER.url} component={DiscoverPage} />
                <PrivateRoute path={ROUTES.UPCOMING.url} component={UpcomingPage} />
                <PrivateRoute path={ROUTES.MY_EVENTS.url} component={MyEventsPage} />
                <PrivateRoute exact path={ROUTES.HOME}>
                  <Redirect to={ROUTES.DISCOVER.url} />
                </PrivateRoute>
              </SwipeableRoutes>
```

This is easily fixed with the changes in this PR. Instead of using SwipeableRoutes, we use a simple Switch. This fixes all the issues I listed. The downsides are that we can't swipe between the home pages, and there's no animation when you transition between the pages. I don't see this as a big deal.

Just a side note, this also ends up fixing the flakiness I had with the e2e for US-2.8, since that was caused by the SwipeableRoutes

### DIFF
--- a/apps/mull-ui/src/app/app.tsx
+++ b/apps/mull-ui/src/app/app.tsx
@@ -3,7 +3,6 @@ import axios from 'axios';
 import jwtDecode from 'jwt-decode';
 import React, { CSSProperties, useEffect, useState } from 'react';
 import { Redirect, Route, Switch, useLocation } from 'react-router-dom';
-import SwipeableRoutes from 'react-swipeable-routes';
 import { toast, ToastContainer } from 'react-toastify';
 import 'react-toastify/dist/ReactToastify.min.css';
 import { ROUTES } from '../constants';
@@ -97,14 +96,14 @@ export const App = () => {
               className="top-nav-bar-shadow"
             />
             <div className="page-container with-sub-nav-bar">
-              <SwipeableRoutes>
+              <Switch>
                 <PrivateRoute path={ROUTES.DISCOVER.url} component={DiscoverPage} />
                 <PrivateRoute path={ROUTES.UPCOMING.url} component={UpcomingPage} />
                 <PrivateRoute path={ROUTES.MY_EVENTS.url} component={MyEventsPage} />
-                <PrivateRoute exact path={ROUTES.HOME}>
+                <PrivateRoute path={ROUTES.HOME}>
                   <Redirect to={ROUTES.DISCOVER.url} />
                 </PrivateRoute>
-              </SwipeableRoutes>
+              </Switch>
             </div>
           </PrivateRoute>
           <PrivateRoute path={ROUTES.MESSAGES} component={MessagesRoute} />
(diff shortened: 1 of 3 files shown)


ys-lin:
#334 also has a solution to the problems mentioned without removing swipeable routes(just a few lines of changes), but if this is going to cause more issues, I think it should be removed

cristian-aldea:
Closing for now, we will go with the solution given in #334

PR #351, opened 2021-04-08
FIX: Create global styling for empty array messages

RGPosadas:
This creates common styling for all empty array messages.

If an original div class had more than `text-align: center` as its styling, then that className was kept and `empty-array-msg` was added on top after removing repetitions. Otherwise, that className was deleted and was replaced with `empty-array-msg`

Before:
image

After:
image


r-saba (approved):
LGTM! Really helps the UI
image

cristian-aldea (approved):

PR #352, opened 2021-04-08
TASK-92: Loading Spinner

### LINKED ISSUE #320: TASK-92: Loading Spinner
**Task Description**

As a developer, I want to implement a loading spinner instead of just using a `Loading...` text div.

**Acceptance Criteria**

bit172:
closes #320 

This has 2 parts: 
  - the Spinner component
  - the withSpinner HOC (Higher-Order Component)
 
Higher-Order Components wrap other components and provide them with extra functionality.
In this case, `withSpinnerHOC` will give the component an extra prop called `isLoading` and it will
conditionally render a spinner or the component based on that value.

Uses:

```jsx
...
return (...
<Spinner />
...)
```

```jsx
// SomeComponent.tsx  ---------------------------------------------------

const SomeComponent = () => {...};
export default withSpinnerHOC(SomeComponent);

//-----------------------------------------------------------------------
// somewhere in the code
let isLoading = true; // this will change over time
return (<SomeComponent isLoading={isLoading} />


```
**Steps to test**

1. Launch the application
2. Open your dev tools and go to the Network tab
3. The should be a throttling options
image
4. Set it to `Slow 3G` to throttle the network so that the spinner appears for longer
5. Reload the application and see at the loading spinners on different pages

### DIFF
--- a/apps/mull-ui/src/app/app.tsx
+++ b/apps/mull-ui/src/app/app.tsx
@@ -65,7 +65,6 @@ export const App = () => {
     return {};
   };
 
-  // TODO: Replace with spinner or loading component
   if (loading) return <Spinner />;
 
   return (
--- a/apps/mull-ui/src/app/app.tsx
+++ b/apps/mull-ui/src/app/app.tsx
@@ -11,6 +11,7 @@ import { environment } from '../environments/environment';
 import { setAccessToken } from './access-token';
 import './app.scss';
 import { BotNavBar, PrivateRoute, SubNavBar, TopNavBar } from './components';
+import Spinner from './components/spinner/spinner';
 import { UserProvider } from './context/user.context';
 import NotFoundPage from './pages/404/not-found-page';
 import CreateEventPage from './pages/create-event/create-event';
@@ -65,7 +66,7 @@ export const App = () => {
   };
 
   // TODO: Replace with spinner or loading component
-  if (loading) return <div>Loading...</div>;
+  if (loading) return <Spinner />;
 
   return (
     <UserProvider value={{ userId, setUserId }}>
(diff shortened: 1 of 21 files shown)


apps/mull-ui/src/app/pages/event-page/event-page.tsx
```diff
@@ -73,12 +74,11 @@ export const EventPage = ({
   }
 
   if (!loadingUser && dataUser) {
-    event.host = dataUser.user;
+    event.host = (dataUser.user as unknown) as IUser;
```
cristian-aldea: It would be better if you set the type to `User` from the generated graphql

```suggestion
    event.host = (dataUser.user as unknown) as User;
```

And set the event to be of type `Event`, again from the generated graphql.

I also fix the typescript error in my pr. have a look at how I do it there

apps/mull-ui/src/app/app.tsx
```diff
@@ -65,7 +66,7 @@ export const App = () => {
   };
 
   // TODO: Replace with spinner or loading component
```
cristian-aldea: You can remove the todo

cristian-aldea (approved):

apps/mull-ui/src/app/pages/event-page/event-page.tsx
bit172: ~~Ah, I guess I'll leave it as it was before and let your PR handle the types. I just got annoyed by the red lines here.~~
Nevermind, I'll add the types since it would cause TS errors on develop.

RGPosadas (approved):
Linting + tests pass. LGTM!

**Acceptance Criteria**

apps/mull-ui/src/app/app.tsx
bit172: This will be done in #352

PR #358, opened 2021-04-08
BUG-29: Direct Message List Does Not Use True Friends

### LINKED ISSUE #356: BUG-29: Direct Message List Does Not Use True Friends
**Describe The Bug**

The direct message list shows pending friends instead of true friends.

**To Reproduce**

1. Go to the messages page
2. See that there are pending friends on the list

**Expected Behavior**

The direct message list should only show true friends.

**Bug Handling Strategy**
- [X] Fix it now
- [ ] Fix it later

bit172:
closes #356 

**Steps to test**
1. Have a pending friend request
2. Go to messages and see that the pending friend is not in the direct message list

### DIFF
--- a/apps/mull-api/src/app/user/user.resolver.ts
+++ b/apps/mull-api/src/app/user/user.resolver.ts
@@ -46,7 +46,7 @@ export class UserResolver {
   }
 
   @Query(/* istanbul ignore next */ () => [Friend])
-  async friends(@AuthenticatedUser() id: number) {
+  async getTrueFriends(@AuthenticatedUser() id: number) {
     const friends = this.filterTrueFriends(id, await this.userService.getFriends(id));
     const modifiedFriends: Friend[] = [];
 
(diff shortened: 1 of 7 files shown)


apps/mull-api/src/app/user/user.resolver.ts
```diff
@@ -37,7 +47,7 @@ export class UserResolver {
 
   @Query(/* istanbul ignore next */ () => [Friend])
   async friends(@AuthenticatedUser() id: number) {
-    const friends = await this.userService.getFriends(id);
+    const friends = this.filterTrueFriends(id, await this.userService.getFriends(id));
```
cristian-aldea, line 50: `friends` on purpose takes only the friends you have in your friend list and doesn't check for 2 way btw. You should change your call in the frontend from `friends` to `getTrueFriends`

cristian-aldea (commented):
I don't think this is the fix that should be done. You should change the frontend call, not the backend one, please.

ys-lin (approved):
Works great, LGTM :)

cristian-aldea (approved):

JimLau49 (approved):
Working good LGTM

RGPosadas (approved):

PR #361, opened 2021-04-09
FIX: Fix Failing E2Es

RGPosadas:
How to test:
* Ensure all of the E2Es pass

Important notes:
* You need to have a f**resh new database that is seeded** before conducting any of the E2Es
* 2.7 and 2.8 are known to be flaky
* 8.1 and 8.5 will run once successfully. If you want to set them up again, then follow the friends relations in[ 8.1's setup](https://github.com/RGPosadas/Mull/blob/develop/apps/mull-api/src/database/seeders/database.seeder.ts#L60) and [8.5's setup](https://github.com/RGPosadas/Mull/blob/develop/apps/mull-api/src/database/seeders/database.seeder.ts#L64).

image

### DIFF
--- a/apps/mull-ui-e2e/src/integration/US-1.1.spec.ts
+++ b/apps/mull-ui-e2e/src/integration/US-1.1.spec.ts
@@ -108,10 +108,10 @@ frameSizes.forEach((frame) => {
     it('should change restriction open', () => {
       cy.get('[data-testid=pill-id-1]')
         .click()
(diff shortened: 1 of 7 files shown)


cindyslittleplanet (approved):
Except the flaky 2.7 and 2.8 E2E which failed for me, every other E2E passes, so LGTM and thank you for fixing any broken E2Es we missed during our PRs. :heart:

cristian-aldea (approved):

PR #367, opened 2021-04-10
SONARCLOUD-Sprint9

cindyslittleplanet:
This PR addresses issues brought up by SonarCloud.

- Add the entire `mockdata` folder to exclusions because it kept bringing up false security hotspots.
- Add more folders to the exclusions because they were irrelevant to the code coverage.
- Remove the empty source `SCSS` files.
- Remove redundant `await` on a non-promise in `event.resolver.ts`.
- Return the query immediately rather than using the temporary variable `strangers` in `user.service.ts`.
- Remove unused declaration of `const post` in `direct-message-chat.tsx`.


ys-lin (approved):
LGTM :D

bit172 (approved):
LGTM! 🚀🚀

